PR #28, opened 2021-11-05
Add Process Instance assertions

### LINKED ISSUE #1: Add ProcessInstance assertions
**Description**

We would like to have assertions that enable us to assert specific conditions of a Process Instance. Examples of some assertions could be:

- Verify a process instance has been started
- Verify a process instance has been completed
- Verify a process instance has been terminated

**References**
[Camunda Platform PI assertions](https://github.com/camunda/camunda-bpm-assert/blob/master/core/src/main/java/org/camunda/bpm/engine/test/assertions/bpmn/ProcessInstanceAssert.java)

remcowesterhoud:
## Description

- Some initial setup
- Added Proces Instance assertions
- Added a StreamFilter used for filtering the Record Stream
- Fixed the dependency hell we got by depending on EZE
- Fixed the CI to use Java 11

## Related issues

closes #1 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/main/java/io/camunda/testing/assertions/BpmnAssertions.java
+++ b/src/main/java/io/camunda/testing/assertions/BpmnAssertions.java
@@ -1,0 +1,12 @@
+package io.camunda.testing.assertions;
+
+import org.camunda.community.eze.RecordStreamSource;
+
+public abstract class BpmnAssertions {
+
+  static ThreadLocal<RecordStreamSource> recordStreamSource = new ThreadLocal<>();
+
+  public static void init(RecordStreamSource recordStreamSource) {
+    BpmnAssertions.recordStreamSource.set(recordStreamSource);
+  }
+}
--- a/src/main/java/io/camunda/testing/assertions/BpmnAssertions.java
+++ b/src/main/java/io/camunda/testing/assertions/BpmnAssertions.java
@@ -1,5 +1,6 @@
 package io.camunda.testing.assertions;
 
+import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
 import org.camunda.community.eze.RecordStreamSource;
 
 public abstract class BpmnAssertions {
@@ -8,5 +9,18 @@ public abstract class BpmnAssertions {
 
   public static void init(RecordStreamSource recordStreamSource) {
     BpmnAssertions.recordStreamSource.set(recordStreamSource);
+  }
+
+  public static ProcessInstanceAssertions assertThat(final ProcessInstanceEvent instanceEvent) {
+    return new ProcessInstanceAssertions(instanceEvent, getRecordStreamSource());
+  }
+
+  private static RecordStreamSource getRecordStreamSource() {
+    if (recordStreamSource.get() == null) {
+      throw new AssertionError(
+          "No RecordStreamSource is set. Please use @ZeebeAssertions extension and "
+              + "declare a RecordStreamSource field");
+    }
+    return recordStreamSource.get();
   }
 }
--- a/src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
+++ b/src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
@@ -46,9 +46,9 @@ public class ProcessInstanceAssertions
             .withBpmnElementType(BpmnElementType.PROCESS).stream()
             .findFirst();
 
-    if (!processInstanceRecord.isPresent()) {
-      failWithMessage("Process with key %s was not started", actual.getProcessInstanceKey());
-    }
+    assertThat(processInstanceRecord.isPresent())
+        .withFailMessage("Process with key %s was not started", actual.getProcessInstanceKey())
+        .isTrue();
 
     return this;
   }
@@ -68,9 +68,9 @@ public class ProcessInstanceAssertions
                     record.getIntent() == ProcessInstanceIntent.ELEMENT_COMPLETED
                         || record.getIntent() == ProcessInstanceIntent.ELEMENT_TERMINATED);
 
-    if (!isActive) {
-      failWithMessage("Process with key %s is not active", actual.getProcessInstanceKey());
-    }
+    assertThat(isActive)
+        .withFailMessage("Process with key %s is not active", actual.getProcessInstanceKey())
+        .isTrue();
 
     return this;
   }
@@ -88,9 +88,9 @@ public class ProcessInstanceAssertions
             .withIntent(ProcessInstanceIntent.ELEMENT_COMPLETED).stream()
             .findFirst();
 
-    if (!processInstanceRecord.isPresent()) {
-      failWithMessage("Process with key %s was not completed", actual.getProcessInstanceKey());
-    }
+    assertThat(processInstanceRecord.isPresent())
+        .withFailMessage("Process with key %s was not completed", actual.getProcessInstanceKey())
+        .isTrue();
 
     return this;
   }
@@ -108,9 +108,9 @@ public class ProcessInstanceAssertions
             .withIntent(ProcessInstanceIntent.ELEMENT_COMPLETED).stream()
             .findFirst();
 
-    if (processInstanceRecord.isPresent()) {
-      failWithMessage("Process with key %s was completed", actual.getProcessInstanceKey());
-    }
+    assertThat(processInstanceRecord.isPresent())
+        .withFailMessage("Process with key %s was completed", actual.getProcessInstanceKey())
+        .isFalse();
 
     return this;
   }
@@ -128,9 +128,9 @@ public class ProcessInstanceAssertions
             .withIntent(ProcessInstanceIntent.ELEMENT_TERMINATED).stream()
             .findFirst();
 
-    if (!processInstanceRecord.isPresent()) {
-      failWithMessage("Process with key %s was not terminated", actual.getProcessInstanceKey());
-    }
+    assertThat(processInstanceRecord.isPresent())
+        .withFailMessage("Process with key %s was not terminated", actual.getProcessInstanceKey())
+        .isTrue();
 
     return this;
   }
@@ -148,9 +148,9 @@ public class ProcessInstanceAssertions
             .withIntent(ProcessInstanceIntent.ELEMENT_TERMINATED).stream()
             .findFirst();
 
-    if (processInstanceRecord.isPresent()) {
-      failWithMessage("Process with key %s was terminated", actual.getProcessInstanceKey());
-    }
+    assertThat(processInstanceRecord.isPresent())
+        .withFailMessage("Process with key %s was terminated", actual.getProcessInstanceKey())
+        .isFalse();
 
     return this;
   }
@@ -192,10 +192,9 @@ public class ProcessInstanceAssertions
             .withIntent(ProcessInstanceIntent.ELEMENT_COMPLETED).stream()
             .count();
 
-    if (count != times) {
-      failWithActualExpectedAndMessage(
-          count, times, "Expected element with id %s to be passed %s times", elementId, times);
-    }
+    assertThat(count)
+        .withFailMessage("Expected element with id %s to be passed %s times", elementId, times)
+        .isEqualTo(times);
 
     return this;
   }
@@ -215,7 +214,9 @@ public class ProcessInstanceAssertions
             .map(ProcessInstanceRecordValue::getElementId)
             .collect(Collectors.toList());
 
-    assertThat(foundElementRecords).isEqualTo(Arrays.asList(elementIds));
+    assertThat(foundElementRecords)
+        .describedAs("Ordered elements")
+        .isEqualTo(Arrays.asList(elementIds));
 
     return this;
   }
@@ -233,13 +234,12 @@ public class ProcessInstanceAssertions
             .filter(((Predicate<String>) this::isWaitingAtElement).negate())
             .collect(Collectors.toList());
 
-    if (!notActivatedElements.isEmpty()) {
-      final String errorMessage =
-          String.format(
-              "Process with key %s is not waiting at element(s) with id(s) %s",
-              actual.getProcessInstanceKey(), String.join(", ", notActivatedElements));
-      failWithMessage(errorMessage);
-    }
+    assertThat(notActivatedElements.isEmpty())
+        .withFailMessage(
+            "Process with key %s is not waiting at element(s) with id(s) %s",
+            actual.getProcessInstanceKey(), String.join(", ", notActivatedElements))
+        .isTrue();
+
     return this;
   }
 
@@ -254,24 +254,23 @@ public class ProcessInstanceAssertions
     final List<String> activatedElements =
         Arrays.stream(elementIds).filter(this::isWaitingAtElement).collect(Collectors.toList());
 
-    if (!activatedElements.isEmpty()) {
-      final String errorMessage =
-          String.format(
-              "Process with key %s is waiting at element(s) with id(s) %s",
-              actual.getProcessInstanceKey(), String.join(", ", activatedElements));
-      failWithMessage(errorMessage);
-    }
+    assertThat(activatedElements.isEmpty())
+        .withFailMessage(
+            "Process with key %s is waiting at element(s) with id(s) %s",
+            actual.getProcessInstanceKey(), String.join(", ", activatedElements))
+        .isTrue();
+
     return this;
   }
 
-  // TODO if waiting at any element with this id returns true. Maybe add a counter so we can check
-  // we are waiting at multiple elements with the same id
   /**
    * Checks if the process instance is currently waiting at an element with a specific element id.
    *
    * @param elementId The id of the element
    * @return boolean indicating whether the process instance is waiting at the element
    */
+  // TODO if waiting at any element with this id returns true. Maybe add a counter so we can check
+  // we are waiting at multiple elements with the same id
   private boolean isWaitingAtElement(final String elementId) {
     final Map<Long, List<Record<ProcessInstanceRecordValue>>> recordsMap =
         StreamFilter.processInstance(recordStreamSource.processInstanceRecords())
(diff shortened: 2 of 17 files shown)


src/main/java/io/camunda/testing/assertions/BpmnAssertions.java
```diff
@@ -0,0 +1,26 @@
...
+
+import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
+import org.camunda.community.eze.RecordStreamSource;
+
+public abstract class BpmnAssertions {
+
+  static ThreadLocal<RecordStreamSource> recordStreamSource = new ThreadLocal<>();
+
+  public static void init(RecordStreamSource recordStreamSource) {
+    BpmnAssertions.recordStreamSource.set(recordStreamSource);
```
pihme, line 11: :x: I think we should also clear this after a test. Otherwise we accumulate objects on the heap, and/or the next test on the same thread could run against the record stream of the test before it. This would be very hard to detect

src/main/java/io/camunda/testing/filters/ProcessInstanceRecordStreamFilter.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public ProcessInstanceRecordStreamFilter(
+      final Iterable<Record<ProcessInstanceRecordValue>> records) {
+    stream = StreamSupport.stream(records.spliterator(), false);
+  }
+
+  public ProcessInstanceRecordStreamFilter withProcessInstanceKey(final long processInstanceKey) {
+    stream =
+        stream.filter(record -> record.getValue().getProcessInstanceKey() == processInstanceKey);
+    return this;
```
pihme, line 23: :wrench: We could return a new object here, immutable function programming style. Same in the other methods

src/main/java/io/camunda/testing/filters/ProcessMessageSubscriptionRecordStreamFilter.java
```diff
@@ -0,0 +1,32 @@
...
+  public ProcessMessageSubscriptionRecordStreamFilter(
+      final Iterable<Record<ProcessMessageSubscriptionRecordValue>> records) {
+    stream = StreamSupport.stream(records.spliterator(), false);
+  }
+
+  public ProcessMessageSubscriptionRecordStreamFilter withProcessInstanceKey(
+      final long processInstanceKey) {
+    stream =
+        stream.filter(record -> record.getValue().getProcessInstanceKey() == processInstanceKey);
+    return this;
```
pihme, line 21: dito

src/main/java/io/camunda/testing/filters/StreamFilter.java
```diff
@@ -0,0 +1,18 @@
+package io.camunda.testing.filters;
+
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.value.ProcessInstanceRecordValue;
+import io.camunda.zeebe.protocol.record.value.ProcessMessageSubscriptionRecordValue;
+
+public class StreamFilter {
+
+  public static ProcessInstanceRecordStreamFilter processInstance(
```
pihme, line 9: :wrench: I think we could add convenience methods that take `RecordStreamSource`

`StreamFilter.processInstance(recordStreamSource.processInstanceRecords())`-> `StreamFilter.processInstance(recordStreamSource))`

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -0,0 +1,413 @@
...
+  }
+
+  /**
+   * Verifies the expectation that the process instance is started. This will also be true when the
+   * process has been completed or terminated.
+   *
+   * @return this {@link ProcessInstanceAssertions}
+   */
+  public ProcessInstanceAssertions isStarted() {
+    final Optional<Record<ProcessInstanceRecordValue>> processInstanceRecord =
```
pihme, line 43: :x: I think this condition is incorrect. I think you need to look for an ACTIVATED intent.

For example, when you start a process instance and immediately an incident is created, or the command gets rejected, then this method would return true, but the process instance was never started.

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -0,0 +1,413 @@
...
+   */
+  public ProcessInstanceAssertions isNotCompleted() {
+    final Optional<Record<ProcessInstanceRecordValue>> processInstanceRecord =
+        StreamFilter.processInstance(recordStreamSource.processInstanceRecords())
+            .withProcessInstanceKey(actual.getProcessInstanceKey())
+            .withBpmnElementType(BpmnElementType.PROCESS)
+            .withIntent(ProcessInstanceIntent.ELEMENT_COMPLETED).stream()
+            .findFirst();
+
+    assertThat(processInstanceRecord.isPresent())
```
pihme, line 111: :x: 
This could be refactored to:
```
Stream<Record> completedElements = sharedCode();

a) actualCompleted = completedElements.findFirst().isPresent();
b) actualNotCompleted = completedEleemnts.findFirst().isEmpty();
```

Same for other methods below

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -0,0 +1,413 @@
...
+  }
+
+  /**
+   * Verifies the expectation that the process instance has passed an element with a specific
+   * element id exactly one time.
+   *
+   * @param elementId The id of the element
+   * @return this {@link ProcessInstanceAssertions}
+   */
+  public ProcessInstanceAssertions hasPassedElement(final String elementId) {
```
pihme, line 165: :thought_balloon: Maybe rename to `hasPassedExactlyOnce()`

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -0,0 +1,413 @@
...
+   * Verifies the expectation that the process instance has passed the given elements in order.
+   *
+   * @param elementIds The element ids
+   * @return this {@link ProcessInstanceAssertions}
+   */
+  public ProcessInstanceAssertions hasPassedElementInOrder(final String... elementIds) {
+    final List<String> foundElementRecords =
+        StreamFilter.processInstance(recordStreamSource.processInstanceRecords())
+            .withProcessInstanceKey(actual.getProcessInstanceKey()).withElementIdIn(elementIds)
+            .withIntent(ProcessInstanceIntent.ELEMENT_ACTIVATING).stream()
```
pihme, line 212: :x: this is inconsistent with `hasPassedElement` in the former the filter is on `ELEMENT_COMPLETED` here it is `ELEMENT_ACTIVATING`. In my mind, `ELEMENT_COMPLETED` is the better indicator and should also be applied here.

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -0,0 +1,413 @@
...
+
+  /**
+   * Verifies the expectation that the process instance is currently waiting at one or more
+   * specified elements.
+   *
+   * @param elementIds The ids of the elements
+   * @return this {@link ProcessInstanceAssertions}
+   */
+  public ProcessInstanceAssertions isWaitingAtElement(final String... elementIds) {
+    final List<String> notActivatedElements =
```
pihme, line 232: :x: the name is confusing

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -0,0 +1,413 @@
...
+  /**
+   * Checks if the process instance is currently waiting at an element with a specific element id.
+   *
+   * @param elementId The id of the element
+   * @return boolean indicating whether the process instance is waiting at the element
+   */
+  // TODO if waiting at any element with this id returns true. Maybe add a counter so we can check
+  // we are waiting at multiple elements with the same id
+  private boolean isWaitingAtElement(final String elementId) {
+    final Map<Long, List<Record<ProcessInstanceRecordValue>>> recordsMap =
```
pihme, line 275: :thought_balloon: This method is difficult to read. I think it is correct, though.

I wonder whether the first map can be reused for other methods in here as well.

pihme: :thought_balloon: I think we could also discuss this with @npepinpe and @saig0 to make sure we have the same semantics of "waiting" in mind.

pihme: :x: Come to think of it, we generally have to check that the command was not rejected

`record -> record.getRejectionType() == RejectionType.NULL_VAL`

pihme (changes requested):
Incomplete review. This is the first batch of comments.

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
saig0: > make sure we have the same semantics of "waiting" in mind.

I think the term "waiting on an element" is common. So, an element is entered but not yet left. 

It is not precise. In Zeebe, we would say that an element (instance) is activated. Or, it is completed. 
In the docs, we use both "entered" and "activated".

But I guess for non-technical users, the term "waiting" sounds more familiar.

pihme: My question was more related to what do we define as wait state? 

Right now we use "the last record for an element was intent = ELEMENT_ACTIVATED" as filter criterion for wait state.
What about situations where it is "waiting" because an incident was raised? Is that waiting, or is that something else?

saig0: > What about situations where it is "waiting" because an incident was raised? Is that waiting, or is that something else?

Good point. 

In Operate, the element should be marked as active. AFAIK, it is active from `activating` to `completed`. 

In general, it would say that an element reached the wait state when it is `activated`. If an incident occurs before then it is not waiting - because you can't trigger the element (e.g. complete the job, send a message, etc.). Same if an incident occurs after `completing` then the element is already on leave.

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -0,0 +1,413 @@
...
+   */
+  public ProcessInstanceAssertions isWaitingExactlyAtElements(final String... elementIds) {
+    final List<String> shouldBeWaitingAt = Arrays.asList(elementIds);
+    final List<String> wrongfullyWaitingElementIds = new ArrayList<>();
+    final List<String> wrongfullyNotWaitingElementIds = new ArrayList<>();
+
+    StreamFilter.processInstance(recordStreamSource.processInstanceRecords())
+        .withProcessInstanceKey(actual.getProcessInstanceKey())
+        .withoutBpmnElementType(BpmnElementType.PROCESS)
+        .withIntent(ProcessInstanceIntent.ELEMENT_ACTIVATING).stream()
```
pihme, line 311: :x: I think this is inconsistent with `isWaitingAt` due to `ELEMENT_ACTIVATING` vs. `ELEMENT_ACTIVATED`

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -0,0 +1,413 @@
...
+        .map(Record::getValue)
+        .map(ProcessInstanceRecordValue::getElementId)
+        .forEach(
```
pihme, line 314: :x: I think this could be refactored to avoid unnecessary calculation.

* the map of element IDs could be reduced to its unique elements
* `isWaitingAtElement(id)` could be cached
* `shouldBeWaitingAt.contains(id)` could be cached

I think this would also make the logic clearer.

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -0,0 +1,413 @@
...
+              if (shouldBeWaitingAt.contains(id) && !isWaitingAtElement(id)) {
+                // not waiting for the element we want to wait for!
+                wrongfullyNotWaitingElementIds.add(id);
+              } else if (!shouldBeWaitingAt.contains(id) && isWaitingAtElement(id)) {
+                // waiting for an element we don't want to wait for!
+                wrongfullyWaitingElementIds.add(id);
+              }
+            });
+
+    assertThat(wrongfullyWaitingElementIds.isEmpty())
```
pihme, line 325: :x: Please use https://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/SoftAssertions.html

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -0,0 +1,413 @@
...
+
+  /**
+   * Verifies the expectation that the process instance is currently waiting to receive one or more
+   * specified messages.
+   *
+   * @param messageNames Names of the messages
+   * @return this {@link ProcessInstanceAssertions}
+   */
+  public ProcessInstanceAssertions isWaitingForMessage(final String... messageNames) {
+    final List<String> notCreatedProcessMessageSubscriptions =
```
pihme, line 347: :thought_balloon: I find this version of filter logic hard to understand/read.

Maybe it would be easier to understand like this:
```
var messageNamesOfOpenSubscriptions=  getOpenProcessMessageSubscritions().stream()
        .map(record -> record.getValue().getMessageName())
       .collect(toList();;
assertThat(messageNamesOfOpenSubscriptions).containsAll(messageNames);
```

I think the more general method `getOpenProcessMessageSubscriptions()` is the more versatile method and easier to read when you add additional logic to its result. Same for other methods e.g. waiting for elements.

src/test/java/io/camunda/testing/assertions/ProcessInstanceAssertionsTest.java
```diff
@@ -0,0 +1,710 @@
...
+  @Nested
+  class HappyPathTests {
+
+    private RecordStreamSource recordStreamSource;
+
+    @Test
+    public void testProcessInstanceIsStarted() throws InterruptedException {
+      // given
+      deployProcess(PROCESS_INSTANCE_BPMN);
+      final Map<String, Object> variables = Collections.singletonMap("totalLoops", 1);
```
pihme, line 49: :wrench: Maybe use constant for total loops and its value

src/test/java/io/camunda/testing/assertions/ProcessInstanceAssertionsTest.java
```diff
@@ -0,0 +1,710 @@
...
+      final ProcessInstanceEvent instanceEvent =
+          startProcessInstance(PROCESS_INSTANCE_ID, variables);
+
+      // when
+      completeTask(ELEMENT_ID);
+
+      // then
+      assertThat(instanceEvent).hasPassedElementInOrder("startevent", ELEMENT_ID, "endevent");
```
pihme, line 191: :thought_balloon: Maybe also use constants (dito for "servicetask1")

src/test/java/io/camunda/testing/assertions/ProcessInstanceAssertionsTest.java
```diff
@@ -0,0 +1,710 @@
...
+  private void completeTask(final String elementId) throws InterruptedException {
+    Thread.sleep(100);
+    Record<JobRecordValue> lastRecord = null;
+    for (Record<JobRecordValue> record : engine.jobRecords().withElementId(elementId)) {
+      if (record.getIntent().equals(JobIntent.CREATED)) {
+        lastRecord = record;
+      }
+    }
+    if (lastRecord != null) {
+      client.newCompleteCommand(lastRecord.getKey()).send().join();
```
pihme, line 695: :clown_face: I am surprised that this works at all, tbh. Not sure if this is a bug or a feature :thinking:

pihme (changes requested):
Full review

pihme:
BTW: feel free to merge parts or all of this into main, if that makes it easier for you to work with other branches. I didn't expect the size of the review, to be honest :thinking:

remcowesterhoud:
I will merge this and process the review in new PR's. This one got a bit out of hand, sorry @pihme. Next ones will be much smaller!

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
remcowesterhoud: I prefer this name and feel like mentioning it in the javadoc would be sufficient. The reason I've added `hasPassedElement` and `hasNotPassedElement` is because it makes the unit test more readable in my opinion. If we choose to rename it I'd rather remove it entirely and encourage people to use `hasPassedElement(elementId, times)`.

src/test/java/io/camunda/testing/assertions/ProcessInstanceAssertionsTest.java
remcowesterhoud: ❓ Why would this not work? I send a complete command with a jobkey. I'm not sure what the bug would be.

pihme: Maybe I am missing something. Who activated the job? To me it looks like this is completing a job before it has been activated.

remcowesterhoud: Ooh right. @Zelldon at some point mentioned to me that it is not required to activate a job before completing it. I'm not sure if it is intended or not.

pihme: Good to know :-)

PR #33, opened 2021-11-08
Add variable assertions

### LINKED ISSUE #4: Add Variable assertions
**Description**

We would like to have assertions that enable us to assert specific conditions of a Variable. Examples of some assertions could be:

- Verify variable has been created
- Verify variable has specific value
- Verify variable has been updated

remcowesterhoud:
## Description

This change makes it possible to assert against the variables of a process. 

## Related issues

closes #4 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
+++ b/src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
@@ -392,6 +392,12 @@ public class ProcessInstanceAssertions
     return openMessageSubscriptions;
   }
 
+  /**
+   * Verifies the process instance has a variable with the specified name
+   *
+   * @param name The name of the variable
+   * @return this ${@link ProcessInstanceAssertions}
+   */
   public ProcessInstanceAssertions hasVariable(final String name) {
     final Map<String, String> variables = getProcessInstanceVariables();
     return assertVariableInMapOfVariables(name, variables);
@@ -418,6 +424,13 @@ public class ProcessInstanceAssertions
     return this;
   }
 
+  /**
+   * Verifies the process instance has a variable with a specific value.
+   *
+   * @param name The name of the variable
+   * @param value The value of the variable
+   * @return this ${@link ProcessInstanceAssertions}
+   */
   public ProcessInstanceAssertions hasVariableWithValue(final String name, final String value) {
     final ZeebeObjectMapper mapper = new ZeebeObjectMapper();
     final String mappedValue = mapper.toJson(value);
@@ -426,13 +439,20 @@ public class ProcessInstanceAssertions
     assertVariableInMapOfVariables(name, variables);
     assertThat(variables)
         .withFailMessage(
-            "Variable '%s' does not have value '%s', as expected, but instead has the value '%s'",
-            name, value, variables.get(name))
+            "The variable '%s' does not have the expected value. The value passed in"
+                + " ('%s') is internally mapped to a JSON String that yields '%s'. However, the "
+                + "actual value (as JSON String) is '%s.",
+            name, value, mappedValue, variables.get(name))
         .containsEntry(name, mappedValue);
 
     return this;
   }
 
+  /**
+   * Returns a Map of variables that belong to this process instance
+   *
+   * @return map of variables
+   */
   private Map<String, String> getProcessInstanceVariables() {
     return StreamFilter.variable(recordStreamSource)
         .withProcessInstanceKey(actual.getProcessInstanceKey())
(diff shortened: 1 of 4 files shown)


src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -388,4 +391,34 @@ public ProcessInstanceAssertions isNotWaitingForMessage(final String... messageN
...
+  public ProcessInstanceAssertions hasVariable(final String name) {
+    extractVariables()
+        .withFailMessage(
+            "Process with key %s does not contain variable with name %s",
+            actual.getProcessInstanceKey(), name)
+        .containsKey(name);
+    return this;
+  }
+
+  public ProcessInstanceAssertions hasVariableWithValue(final String name, final String value) {
```
pihme, line 404: :x: the second parameter needs to be on Object. We need to convert the object value to a JSON string (ideally in exactly the same way that Zeebe would) and then compare the two strings

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -388,4 +391,34 @@ public ProcessInstanceAssertions isNotWaitingForMessage(final String... messageN
...
+    extractVariables()
+        .withFailMessage(
+            "Process with key %s does not contain variable with name %s and value %s",
```
pihme, line 407: :x: I think this error message is not very helpful. It covers two conditions:
* no variable with that name exists
* a variable does exist, but has a different value

For both I think we should offer dedicated error messages. 
In the first case we could print out the variable names we found (maybe it was just a typo)
In the second case we should print out the value we encountered.

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
pihme: You can try an example where the type is a String. Then you need to assert `hasVariableWithValue("test", "\"value\"")` which is not intuitive.

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -388,4 +391,34 @@ public ProcessInstanceAssertions isNotWaitingForMessage(final String... messageN
...
+  }
+
+  public MapAssert<String, String> extractVariables() {
+    final Map<String, String> variables =
+        StreamFilter.variable(recordStreamSource)
+            .withProcessInstanceKey(actual.getProcessInstanceKey())
+            .withRejectionType(RejectionType.NULL_VAL)
+            .stream()
+            .map(Record::getValue)
+            .collect(Collectors.toMap(VariableRecordValue::getName, VariableRecordValue::getValue));
```
pihme, line 420: :x: Here we also need to unmarshall the JSON value to an object

pihme (changes requested):
Looks good!

The devil is in the detail. The value is a JSON string, which is a bit difficult to work with. I think we need to do more here. Not sure whether you want to tackle it immediately or create a follow up issue. But in my mind this will be mandatory for MVP.

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
pihme: `ZeebeObjectMapper`will do most of the work

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
pihme: This may be a little difficult, because we don't know what object to expect. I wouldn't mind dropping this method for now and adding it after MVP if people request it.

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
pihme: Good point!

What about this: "The variable x does not have the expected value. The value passed in 'rawValue' is internally converted to a JSON String that yields 'jsonEquivalent'. However, the actual value (as JSON String) is 'actualJsonValue'."

I know it's a bit tedious, but this conversion but has caused me a great deal of frustration in the past. So anything we can do here to help the user is well warranted, in my mind.

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -4,6 +4,7 @@
 
 import io.camunda.testing.filters.StreamFilter;
 import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
+import io.camunda.zeebe.client.impl.ZeebeObjectMapper;
```
pihme, line 7: :x: Nothing here, but please reword typo in commit message

remcowesterhoud: Looks like some dutch slipped through 😄 
I've edited the message

pihme (approved):
Good work! I think the extra effort on the better error messages will pay off.

PR #34, opened 2021-11-08
Add automatic formatting of pom files

remcowesterhoud:
## Description

Added the spotless maven plugin. This plugin enables us to automatically format our code. I've enabled the google Java formatting and the pom formatting from sortPom

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
+++ b/src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
@@ -43,7 +43,8 @@ public class ProcessInstanceAssertions
     final Optional<Record<ProcessInstanceRecordValue>> processInstanceRecord =
         StreamFilter.processInstance(recordStreamSource.processInstanceRecords())
             .withProcessInstanceKey(actual.getProcessInstanceKey())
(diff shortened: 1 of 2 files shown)


src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
```diff
@@ -43,7 +43,8 @@ public ProcessInstanceAssertions isStarted() {
     final Optional<Record<ProcessInstanceRecordValue>> processInstanceRecord =
```
pihme, line 43: Can you exclude this file from the PR?

pihme (approved):
I think a file slipped through. Once fixed, feel free to merge

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssertions.java
remcowesterhoud: The changes in this file were part of the google java formatting. I think the new plugin has different rules from the one we used before. I do agree with these changes though, so If you don't mind I'd like to merge it with these in them. All the changes are is add new lines at chained method calls.

This is actually how I wrote it but the other plugin kept undoing it 😄

Issue #2, opened 2021-10-14
Add Deployment assertions


remcowesterhoud:
**Description**

We would like to have assertions that enable us to assert specific conditions of a Deployment. Examples of some assertions could be:

- Verify a process has been deployed
- Verify the version of a deployed process 
- Verify the reason why a deployment has failed

pihme:
This was implemented in two assertions:

* `DeploymentAssert` checks the deployment and allows to navigate to `ProcessAssert`
* `ProcessAssert` checks the process (e.g. version, resource name) and can also tell you whether the process has any instances

`ProcessAssert` is inspired by `ProcessDefinitionAssert` in BPMN Assert (https://github.com/camunda/camunda-bpm-assert/blob/master/core/src/main/java/org/camunda/bpm/engine/test/assertions/bpmn/ProcessDefinitionAssert.java)

Deviattions from BPMN assert:
* in BPMN assert you can assert the number of active instances, in our version you (currently) can only assert over the number of instances (active or not)

I think we could have a follow up issue to assert
* active instances
* terminated instances
* Completed instances

And so on. 

For now, I have limited the scope to the bare minimum.

See also #46 for additional ideas to expand the `ProcessAssert

PR #45, opened 2021-11-09
refactor: extract utility logic for tests

pihme:
## Description

We already talked about refactoring the logic we use in the tests. I started doing this for #2 and the realized, I would like to have it in main earlier, and extracted this smaller PR.

The content is a bit narrow and motivated by what I am going to need in #2.

I tried modifying the existing test, but this got messy quite quickly. 

So, since there really is no perfect time, this PR serves as a start for having this class in the code base.

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/test/java/io/camunda/testing/util/Utilities.java
+++ b/src/test/java/io/camunda/testing/util/Utilities.java
@@ -1,0 +1,63 @@
+package io.camunda.testing.util;
+
+import io.camunda.zeebe.client.ZeebeClient;
+import io.camunda.zeebe.client.api.command.DeployProcessCommandStep1;
+import io.camunda.zeebe.client.api.response.DeploymentEvent;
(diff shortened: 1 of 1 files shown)


remcowesterhoud (approved):
⭐ I like this change! It is now not used in any of the tests, is that on purpose or did you forget to commit something?

Either way this code looks good

pihme:
> star I like this change! It is now not used in any of the tests, is that on purpose or did you forget to commit something?

It will be used in #2. I wanted to keep the change set small.

pihme:
bors merge

remcowesterhoud:
@pihme I don't think we have bors in this repo yet

pihme:
> @pihme I don't think we have bors in this repo yet

Old habits :-)

PR #55, opened 2021-11-11
Incident assertions

### LINKED ISSUE #11: Add Incident assertions
**Description**

We would like to have assertions that enable us to assert specific conditions of an Incident. Examples of some assertions could be:

- Incident was created
- Incident has content
- Incident was resolved

pihme:
## Description

Adds incident assertions 

## Related issues

closes #11

## Definition of Done
Code changes:

Testing:
* [X] There are unit/integration tests that verify all acceptance criterias of the issue
* [X] New tests are written to ensure backwards compatibility with further versions

Documentation:

### DIFF
--- a/src/main/java/io/camunda/testing/assertions/IncidentAssert.java
+++ b/src/main/java/io/camunda/testing/assertions/IncidentAssert.java
@@ -1,0 +1,15 @@
+package io.camunda.testing.assertions;
+
+import org.assertj.core.api.AbstractAssert;
+import org.camunda.community.eze.RecordStreamSource;
+
+/** Assertions for incidents. An incident is identified by its incident key */
+public class IncidentAssert extends AbstractAssert<IncidentAssert, Long> {
+
+  private final RecordStreamSource recordStreamSource;
+
+  public IncidentAssert(final long incidentKey, final RecordStreamSource recordStreamSource) {
+    super(incidentKey, IncidentAssert.class);
+    this.recordStreamSource = recordStreamSource;
+  }
+}
--- a/src/main/java/io/camunda/testing/assertions/IncidentAssert.java
+++ b/src/main/java/io/camunda/testing/assertions/IncidentAssert.java
@@ -1,9 +1,22 @@
 package io.camunda.testing.assertions;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.camunda.testing.filters.IncidentRecordStreamFilter;
+import io.camunda.testing.filters.StreamFilter;
+import io.camunda.zeebe.client.api.response.ActivatedJob;
+import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.RejectionType;
+import io.camunda.zeebe.protocol.record.intent.IncidentIntent;
+import io.camunda.zeebe.protocol.record.value.ErrorType;
+import io.camunda.zeebe.protocol.record.value.IncidentRecordValue;
+import java.util.Optional;
 import org.assertj.core.api.AbstractAssert;
+import org.assertj.core.api.StringAssert;
 import org.camunda.community.eze.RecordStreamSource;
 
-/** Assertions for incidents. An incident is identified by its incident key */
+/** Assertions for incidents. An incident is identified by its incident key. */
 public class IncidentAssert extends AbstractAssert<IncidentAssert, Long> {
 
   private final RecordStreamSource recordStreamSource;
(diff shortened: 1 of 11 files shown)


src/test/java/io/camunda/testing/assertions/IncidentAssertTest.java
```diff
@@ -0,0 +1,449 @@
...
+          .join();
+
+      final IncidentAssert incidentAssert = assertThat(actual).extractLatestIncident();
+
+      // then
+      incidentAssert.isUnresolved();
+    }
+  }
+
+  private ActivateJobsResponse activateSingleJob() {
```
remcowesterhoud, line 242: ❌  This could be moved to the testing Utilities. The same method is also used in `JobAssertTest` so it'd be nice to move this to a place where it can be used by both tests classes.

src/test/java/io/camunda/testing/assertions/IncidentAssertTest.java
```diff
@@ -0,0 +1,449 @@
...
+
+      final ActivatedJob actual = jobActivationResponse.getJobs().get(0);
+      client
+          .newThrowErrorCommand(actual.getKey())
+          .errorCode(ERROR_CODE)
+          .errorMessage(ERROR_MESSAGE)
+          .send()
+          .join();
+
+      final IncidentAssert incidentAssert = assertThat(actual).extractLatestIncident();
```
remcowesterhoud, line 57: 🔧 Instead of creating a variable we could chain the actual assertion. Idem for the other tests.

src/main/java/io/camunda/testing/assertions/IncidentAssert.java
```diff
@@ -0,0 +1,214 @@
...
+    return this;
+  }
+
+  /**
+   * Asserts that the incident has the given error message
+   *
+   * @param expectedErrorMessage expected error message
+   * @return this {@link IncidentAssert}
+   */
+  public IncidentAssert hasErrorMessage(final String expectedErrorMessage) {
```
remcowesterhoud, line 62: ❌  It would be cool if we could support string formatting in this assertion. Like AssertJ does in for example the `withFailMessage` below.

I put it as a blocker because I think we should have this before our first release. But I wouldn't mind if it is added at a later stage.

src/test/java/io/camunda/testing/assertions/IncidentAssertTest.java
```diff
@@ -0,0 +1,449 @@
...
+
+      // then
+      incidentAssert.hasErrorMessage(
+          "Expected to throw an error event with the code 'error' with message 'error occurred', but it was not caught. No error events are available in the scope.");
```
remcowesterhoud, line 84: 💭 Should you implement the support for string formatting I mentioned earlier we could use it here for the error code and message.

src/test/java/io/camunda/testing/assertions/IncidentAssertTest.java
```diff
@@ -0,0 +1,449 @@
...
+          .errorMessage(ERROR_MESSAGE)
+          .send()
+          .join();
+
+      final IncidentAssert incidentAssert = assertThat(actual).extractLatestIncident();
+
+      final StringAssert messageAssert = incidentAssert.extractErrorMessage();
+
+      // then
+      Assertions.assertThat(messageAssert).isNotNull();
```
remcowesterhoud, line 109: 🙃 `extractErrorMessage()` will never return null

src/test/java/io/camunda/testing/assertions/IncidentAssertTest.java
```diff
@@ -0,0 +1,449 @@
...
+      // when
+      final ProcessInstanceEvent instanceEvent =
+          startProcessInstance(client, ProcessPackLoopingServiceTask.PROCESS_ID, variables);
+      /* will raise an incident in the gateway because VAR_TOTAL_LOOPS is a string, but needs to be an int */
+      final ActivateJobsResponse jobActivationResponse = activateSingleJob();
+      final ActivatedJob job = jobActivationResponse.getJobs().get(0);
+      client.newCompleteCommand(job.getKey()).send().join();
+
+      // TODO remove this
+      Thread.sleep(100);
```
remcowesterhoud, line 354: 🔧 This construction happens a lot in these tests. It might be nice to extract this to a separate method, or even make a method for it in the testing utilities.

src/test/java/io/camunda/testing/assertions/JobAssertTest.java
```diff
@@ -106,6 +108,65 @@ void testHasRetries() throws InterruptedException {
...
+          .newThrowErrorCommand(actual.getKey())
+          .errorCode("error")
+          .errorMessage("error occurred")
+          .send()
+          .join();
+
+      final IncidentAssert incidentAssert = assertThat(actual).extractLatestIncident();
+
+      // then
+      Assertions.assertThat(incidentAssert).isNotNull();
```
remcowesterhoud, line 167: 🙃 It might make more sense to assert the incident key has a sensible value (> 0). The incident assert will never be null if we manage to extract it. The value of the key could be wrong though.

Same for the extract test in `ProcessInstanceAssertTest`.

src/test/java/io/camunda/testing/assertions/JobAssertTest.java
```diff
@@ -106,6 +108,65 @@ void testHasRetries() throws InterruptedException {
...
+      Utilities.deployProcess(client, ProcessPackLoopingServiceTask.RESOURCE_NAME);
+      Utilities.startProcessInstance(client, ProcessPackLoopingServiceTask.PROCESS_ID);
+
+      // when
+      final ActivateJobsResponse jobActivationResponse = activateSingleJob();
+
+      final ActivatedJob actual = jobActivationResponse.getJobs().get(0);
+      client
+          .newThrowErrorCommand(actual.getKey())
+          .errorCode("error")
```
remcowesterhoud, line 159: 🔧 We could also use constants here the same way as in `IncidentAssertTest`. Idem in the other testcases.

remcowesterhoud (changes requested):
🏅 Thanks @pihme. Overall it looks pretty good! Please don't be alarmed by the amount of comments. Most of them are minor suggestions.

src/main/java/io/camunda/testing/assertions/IncidentAssert.java
```diff
@@ -0,0 +1,214 @@
...
+    return actual;
+  }
+
+  /**
+   * Asserts that the incident has the given error type
+   *
+   * @param expectedErrorType expected error type
+   * @return this {@link IncidentAssert}
+   */
+  public IncidentAssert hasErrorType(final ErrorType expectedErrorType) {
```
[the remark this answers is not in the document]
pihme: Did not implement a check for `expectedErrorMessage` as I think this can be null or empty.

src/main/java/io/camunda/testing/assertions/JobAssert.java
```diff
@@ -84,6 +95,60 @@ public JobAssert hasRetries(final int expectedRetries) {
     return this;
   }
 
+  /**
+   * Asserts whether any incidents were raised for this job (regardless of whether these incidents
+   * are active or already resolved)
+   *
+   * @return this {@link JobAssert}
+   */
+  public JobAssert hasAnyIncidents() {
```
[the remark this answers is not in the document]
pihme: Personally, I like it this way.

So then I can write: `assertThat(job).hasNoIncidents()` which in my opinion is nicer to read than `asserThat(job).hasIncidents(false)`

src/test/java/io/camunda/testing/assertions/IncidentAssertTest.java
pihme: I like it better this way, but not by much. So happy to discuss this if you disagree

src/main/java/io/camunda/testing/assertions/IncidentAssert.java
pihme: Not sure what the benefit is. 

For me `hasErrorMessage("Unexpected Exception: %s" , exceptionMessage);` and `hasErrorMessage(String.format("Unexpected Exception: %s", exceptionMessage));` is kind of the same thing.

And of these two, I like the second better because then I know which syntax to use for placeholders in the string.

src/test/java/io/camunda/testing/assertions/IncidentAssertTest.java
pihme: Well, you know that because you looked in the implementation. But the method could return null.

remcowesterhoud (approved):
Thanks for the changes @pihme! Looks good to merge

PR #56, opened 2021-11-11
Add Message assertions

### LINKED ISSUE #9: Add Message assertions
**Description**

We would like to have assertions that enable us to assert specific conditions of a Message. Examples of some assertions could be:

- Verify Message was published
- Verify Message was correlated
- Verify Message has variables
- Verify Message has time to live

remcowesterhoud:
## Description

- Added Message assertions
- Conntected Message assertions to Process Instance assertions

## Related issues

closes #9 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/main/java/io/camunda/testing/assertions/MessageAssert.java
+++ b/src/main/java/io/camunda/testing/assertions/MessageAssert.java
@@ -7,6 +7,7 @@ import io.camunda.zeebe.client.api.response.PublishMessageResponse;
 import io.camunda.zeebe.protocol.record.Record;
 import io.camunda.zeebe.protocol.record.RejectionType;
 import io.camunda.zeebe.protocol.record.intent.MessageIntent;
+import io.camunda.zeebe.protocol.record.intent.MessageStartEventSubscriptionIntent;
 import io.camunda.zeebe.protocol.record.intent.ProcessMessageSubscriptionIntent;
 import io.camunda.zeebe.protocol.record.value.ProcessMessageSubscriptionRecordValue;
 import java.util.List;
@@ -119,15 +120,17 @@ public class MessageAssert extends AbstractAssert<MessageAssert, PublishMessageR
     return this;
   }
 
-  public ProcessInstanceAssert extractingProcessInstance() {
+  /**
+   * Extracts the process instance for the given message
+   *
+   * @param isMessageStartEvent is the message a start event
+   * @return Process instance assertions {@link ProcessInstanceAssert}
+   */
+  public ProcessInstanceAssert extractingProcessInstance(final boolean isMessageStartEvent) {
     final List<Long> correlatedProcessInstances =
-        StreamFilter.processMessageSubscription(recordStreamSource)
-            .withMessageKey(actual.getMessageKey())
-            .withRejectionType(RejectionType.NULL_VAL)
-            .withIntent(ProcessMessageSubscriptionIntent.CORRELATED)
-            .stream()
-            .map(record -> record.getValue().getProcessInstanceKey())
-            .collect(Collectors.toList());
+        isMessageStartEvent
+            ? getProcessInstanceKeysForCorrelatedMessageStartEvent()
+            : getProcessInstanceKeysForCorrelatedMessage();
 
     Assertions.assertThat(correlatedProcessInstances)
         .withFailMessage(
@@ -136,5 +139,35 @@ public class MessageAssert extends AbstractAssert<MessageAssert, PublishMessageR
         .hasSize(1);
 
     return new ProcessInstanceAssert(correlatedProcessInstances.get(0), recordStreamSource);
+  }
+
+  /**
+   * Gets the correlated process instance keys for process message subscriptions
+   *
+   * @return List of process instance keys
+   */
+  private List<Long> getProcessInstanceKeysForCorrelatedMessage() {
+    return StreamFilter.processMessageSubscription(recordStreamSource)
+        .withMessageKey(actual.getMessageKey())
+        .withRejectionType(RejectionType.NULL_VAL)
+        .withIntent(ProcessMessageSubscriptionIntent.CORRELATED)
+        .stream()
+        .map(record -> record.getValue().getProcessInstanceKey())
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Gets the correlated process instance keys for message start event subscriptions
+   *
+   * @return List of process instance keys
+   */
+  private List<Long> getProcessInstanceKeysForCorrelatedMessageStartEvent() {
+    return StreamFilter.messageStartEventSubscription(recordStreamSource)
+        .withMessageKey(actual.getMessageKey())
+        .withRejectionType(RejectionType.NULL_VAL)
+        .withIntent(MessageStartEventSubscriptionIntent.CORRELATED)
+        .stream()
+        .map(record -> record.getValue().getProcessInstanceKey())
+        .collect(Collectors.toList());
   }
 }
--- a/src/main/java/io/camunda/testing/assertions/MessageAssert.java
+++ b/src/main/java/io/camunda/testing/assertions/MessageAssert.java
@@ -93,7 +93,9 @@ public class MessageAssert extends AbstractAssert<MessageAssert, PublishMessageR
             .isPresent();
 
     assertThat(isCorrelated)
-        .withFailMessage("Message with key %d was not correlated", actual.getMessageKey())
+        .withFailMessage(
+            "Message with key %d did not lead to the creation of a process instance",
+            actual.getMessageKey())
         .isTrue();
 
     return this;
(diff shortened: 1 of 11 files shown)


src/main/java/io/camunda/testing/assertions/MessageAssert.java
```diff
@@ -0,0 +1,90 @@
...
+    super(actual, MessageAssert.class);
+    this.recordStreamSource = recordStreamSource;
+  }
+
+  /**
+   * Verifies the expectation that a message has been correlated
+   *
+   * @return this {@link MessageAssert}
+   */
+  public MessageAssert hasBeenCorrelated() {
```
pihme, line 32: :x: please also add inverse assertion, dio for expired

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssert.java
```diff
@@ -461,4 +516,17 @@ public ProcessInstanceAssert hasVariableWithValue(final String name, final Strin
         .map(Record::getValue)
         .collect(Collectors.toMap(VariableRecordValue::getName, VariableRecordValue::getValue));
   }
+
+  protected static class ProcessInstanceActual {
```
pihme, line 520: :question: What is the purpose of this class?

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssert.java
```diff
@@ -392,6 +392,61 @@ public ProcessInstanceAssert isNotWaitingForMessage(final String... messageNames
...
+  /**
+   * Verifies the expectation that a message with a given name has been correlated a given amount of
+   * times.
+   *
+   * @param messageName The name of the message
+   * @param times The expected amount of times the message is correlated
+   * @return this {@link ProcessInstanceAssert}
+   */
+  public ProcessInstanceAssert hasCorrelatedMessageByName(
+      final String messageName, final int times) {
```
pihme, line 404: :wrench: Maybe check that `messageName` is not empty and times is positive; ditor forother method

src/test/java/io/camunda/testing/assertions/MessageAssertTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+      // then
+      assertThat(response).hasBeenCorrelated();
+    }
+
+    @Test
+    void testHasExpired() throws InterruptedException {
+      // given
+      deployProcess(client, ProcessPackMessageEvent.RESOURCE_NAME);
+      final Duration timeToLive = Duration.ofNanos(1);
```
pihme, line 58: :x: Please use bigger timescales

src/test/java/io/camunda/testing/assertions/MessageAssertTest.java
```diff
@@ -0,0 +1,143 @@
...
+      final PublishMessageResponse response =
+          sendMessage(client, ProcessPackMessageEvent.MESSAGE_NAME, CORRELATION_KEY, timeToLive);
+      clock.increaseTime(timeToLive.plusMillis(1));
+
+      // then
+      assertThat(response).hasExpired();
+    }
+
+    @Test
+    void testExtractingProcessInstance() throws InterruptedException {
```
pihme, line 70: :thought_balloon: I think the example is not very helpful to understand the purpose of this method. Would be better to have a process that is started by a message event. Then it would be clearer why this is useful

src/test/java/io/camunda/testing/util/Utilities.java
```diff
@@ -64,4 +73,28 @@ public static ProcessInstanceEvent startProcessInstance(
     Thread.sleep(100);
     return instanceEvent;
   }
+
+  public static PublishMessageResponse sendMessage(
+      final ZeebeClient client, final String messsageName, final String correlationKey)
+      throws InterruptedException {
+    return sendMessage(client, messsageName, correlationKey, Duration.ofDays(99999));
```
pihme, line 80: :thought_balloon: I think we use -1 or some such value to indicate "forever"

pihme (changes requested):
Good stuff, but needs some more refinement

src/main/java/io/camunda/testing/assertions/ProcessInstanceAssert.java
remcowesterhoud: I was overthinking this yesterday. I decided to introduce this class so we could have an entry point to `ProcessInstanceAssert` that wouldn't be a `ProcessInstanceEvent`. Therefore I created this class, but since we only use the process instance key we might as well use the key itself as the actual.

src/test/java/io/camunda/testing/util/Utilities.java
remcowesterhoud: I've tested this and this does not seem to be the case. When I enter -1 days the messages expire instantly.

src/test/java/io/camunda/testing/assertions/MessageAssertTest.java
remcowesterhoud: I have added more tests for testing if a process is started. For this I had to look at a different set of records. I've also added 2 more assertions to verify whether a start event message has been correlated or not.

src/main/java/io/camunda/testing/assertions/MessageAssert.java
```diff
@@ -67,6 +97,28 @@ public MessageAssert hasExpired() {
...
+        StreamFilter.message(recordStreamSource)
+            .withKey(actual.getMessageKey())
+            .withRejectionType(RejectionType.NULL_VAL)
+            .withIntent(MessageIntent.EXPIRED)
+            .stream()
+            .findFirst()
+            .isPresent();
+
+    assertThat(isExpired)
+        .withFailMessage("Message with key %d was expired", actual.getMessageKey())
```
pihme, line 167: `Message with key %d has expired`

pihme: Dito for the positive case. Missed that earlier

src/main/java/io/camunda/testing/assertions/MessageAssert.java
```diff
@@ -119,15 +120,17 @@ public MessageAssert hasNotExpired() {
...
   }
 
-  public ProcessInstanceAssert extractingProcessInstance() {
+  /**
+   * Extracts the process instance for the given message
+   *
+   * @param isMessageStartEvent is the message a start event
+   * @return Process instance assertions {@link ProcessInstanceAssert}
+   */
+  public ProcessInstanceAssert extractingProcessInstance(final boolean isMessageStartEvent) {
```
pihme, line 179: :question: Is there some way we can get rid of the parameter? Can we somehow figure that out by looking at both record types and picking the one that matches?

src/main/java/io/camunda/testing/assertions/MessageAssert.java
```diff
@@ -49,6 +50,28 @@ public MessageAssert hasBeenCorrelated() {
     return this;
   }
 
+  /**
+   * Verifies the expectation that a message start event has been correlated
+   *
+   * @return this {@link MessageAssert}
+   */
+  public MessageAssert hasMessageStartEventBeenCorrelated() {
```
pihme, line 58: :x: this sounds wrong. The message start event is not correlated, the message is.

rename to: `hasCreatedProcessInstance` 
Also adjust error message accordingly

src/main/java/io/camunda/testing/assertions/MessageAssert.java
```diff
@@ -76,6 +99,33 @@ public MessageAssert hasNotBeenCorrelated() {
     return this;
   }
 
+  /**
+   * Verifies the expectation that a message start event has not been correlated
+   *
+   * @return this {@link MessageAssert}˚
+   */
+  public MessageAssert hasMessageStartEventNotBeenCorrelated() {
```
pihme, line 107: same comment

pihme (changes requested):
Great stuff. Still some changes remaining.

src/main/java/io/camunda/testing/assertions/MessageAssert.java
```diff
@@ -99,12 +77,34 @@ public MessageAssert hasNotBeenCorrelated() {
...
+        StreamFilter.messageStartEventSubscription(recordStreamSource)
+            .withMessageKey(actual.getMessageKey())
+            .withRejectionType(RejectionType.NULL_VAL)
+            .withIntent(MessageStartEventSubscriptionIntent.CORRELATED)
+            .stream()
+            .findFirst()
+            .isPresent();
+
+    assertThat(isCorrelated)
+        .withFailMessage("Message with key %d was not correlated", actual.getMessageKey())
```
pihme, line 96: :x: The error message is incorrect or imprecise. The message may have been correlated, but not to a message start event.
Suggestions:
* Message with key %d was not correlated to a message start event
* Message with key %d did not lead to the creation of a process instance.

pihme (approved):
Looks very good!

One last error message to rephrase, but then it's good to go!

Issue #46, opened 2021-11-10
Add means to assert on process instances that were created indirectly


pihme:
**Description**
Right now we have asserts for process instances that we created directly.
```
      final ProcessInstanceEvent instanceEvent =
          startProcessInstance(PROCESS_INSTANCE_ID, variables);

      // then
      assertThat(instanceEvent).hasPassedElement(ELEMENT_ID);
```

What we currently cannot test is process instances that were created indirectly, e.g. via a message start event, a timer start event, via a call activity.

We should add some way to also do asserts on those.

This could be implemented in several ways:
* through deployment `assertThat(deploymentEvent).extractingProcessByBpmnId(id).getFirstCreatedInstance() ...`
* through some kind of inspection utility `InspectionUtility.ofRecords(records).findProcessInstancesById(id).filterTerminated().getFirst()...`

PR #57, opened 2021-11-15
Timer assertions

### LINKED ISSUE #10: Add Timer assertions
**Description**

We would like to have assertions that enable us to assert specific conditions of a Timer. Examples of some assertions could be:

- Verify a Timer was triggered
- Verify a Timer was triggered n times

remcowesterhoud:
## Description

Timers don't need their own assertions. What we do need is a way to do assertions on a Process Instance once a timer start event has been triggered. This change allows for a route to the Process Instance assertions using the timer id.

I went with the option of using an `InspectionUtility`. There is unfortunately no obvious entry point for timers which could be used in `BpmnAssert`. The only thing we could use is a timer id, but an `assertThat(String)` method is in my opinion too generic. Another option would be choosing a different name for the assertion entry method in `BpmnAssert` so that we could use the timer id. It would result in something like: `assertThatProcessInstanceStartedByTimer(String timerId)` or something like that. This would be confusing in my opinion, as it seems like this will just do a single assertion, instead of returning the `ProcessInstanceAssertion` class.

## Related issues

closes #10
relates to #46 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/main/java/io/camunda/testing/utils/InspectionUtility.java
+++ b/src/main/java/io/camunda/testing/utils/InspectionUtility.java
@@ -1,6 +1,6 @@
 package io.camunda.testing.utils;
 
-import static io.camunda.testing.utils.RecordStreamManager.getRecordStreamSource;
+import static io.camunda.testing.utils.RecordStreamSourceStore.getRecordStreamSource;
 
 import io.camunda.testing.filters.StreamFilter;
 
--- a/src/main/java/io/camunda/testing/utils/InspectionUtility.java
+++ b/src/main/java/io/camunda/testing/utils/InspectionUtility.java
@@ -1,0 +1,12 @@
+package io.camunda.testing.utils;
+
+import static io.camunda.testing.utils.RecordStreamManager.getRecordStreamSource;
+
+import io.camunda.testing.filters.StreamFilter;
+
+public class InspectionUtility {
+
+  public static ProcessEventInspections findProcessEvents() {
+    return new ProcessEventInspections(StreamFilter.processEventRecords(getRecordStreamSource()));
+  }
+}
--- a/src/main/java/io/camunda/testing/utils/ProcessEventInspections.java
+++ b/src/main/java/io/camunda/testing/utils/ProcessEventInspections.java
@@ -1,0 +1,59 @@
+package io.camunda.testing.utils;
+
+import static io.camunda.testing.utils.RecordStreamManager.getRecordStreamSource;
+
+import io.camunda.testing.assertions.ProcessInstanceAssert;
+import io.camunda.testing.filters.ProcessEventRecordStreamFilter;
+import io.camunda.zeebe.protocol.record.intent.ProcessEventIntent;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.assertj.core.api.Assertions;
+
+public class ProcessEventInspections {
+
+  private final ProcessEventRecordStreamFilter filter;
+
+  public ProcessEventInspections(final ProcessEventRecordStreamFilter filter) {
+    this.filter = filter;
+  }
+
+  public ProcessEventInspections triggeredByTimer(final String timerId) {
+    return new ProcessEventInspections(
+        filter.withIntent(ProcessEventIntent.TRIGGERED).withTargetElementId(timerId));
+  }
+
+  public ProcessEventInspections withProcessDefinitionKey(final long processDefinitionKey) {
+    return new ProcessEventInspections(filter.withProcessDefinitionKey(processDefinitionKey));
+  }
+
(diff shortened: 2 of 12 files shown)


src/test/java/io/camunda/testing/utils/ProcessEventInspectionsTest.java
```diff
@@ -0,0 +1,92 @@
...
+    public void testAssertThatFirstProcessInstance() throws InterruptedException {
+      // given
+      final DeploymentEvent deploymentEvent =
+          deployProcess(client, ProcessPackTimerStartEvent.RESOURCE_NAME);
+
+      // when
+      increaseTime(clock, Duration.ofDays(1));
+
+      // then
+      findProcessEvents()
```
pihme, line 43: :x: I am not a fan of this API. I would rather like to see something like:

```java
InstanceEvent processInstance = findProcessEvents()
          .triggeredByTimer(ProcessPackTimerStartEvent.TIMER_ID)
          .withProcessDefinitionKey(deploymentEvent.getProcesses().get(0).getProcessDefinitionKey())
          .firstProcessInstance();
          
assertThat(processInstance).isCompleted();
```

src/main/java/io/camunda/testing/utils/RecordStreamManager.java
```diff
@@ -0,0 +1,25 @@
+package io.camunda.testing.utils;
+
+import org.camunda.community.eze.RecordStreamSource;
+
+public abstract class RecordStreamManager {
```
pihme, line 5: :wrench: Maybe rename to `RecordStreanStore` I am not sure about calling it manager.

src/test/java/io/camunda/testing/utils/ProcessEventInspectionsTest.java
pihme: So then the separation of concerns would be:
* Inspection utility is used to find an entity, that cannot be accessed in a different way
* BPMN assert then does assertions on that entity

Btw, I don't think the result of `firstProcessInstance()` needs to be an `InstanceEvent` if that is inconvenient. It could also return an `InspectedProcessInstance` and then have a different entry point in BPMNAssert

src/main/java/io/camunda/testing/utils/InspectionUtility.java
```diff
@@ -0,0 +1,12 @@
+package io.camunda.testing.utils;
+
+import static io.camunda.testing.utils.RecordStreamManager.getRecordStreamSource;
+
+import io.camunda.testing.filters.StreamFilter;
+
+public class InspectionUtility {
```
pihme, line 7: :x: Please add JavaDocs for class and method

src/main/java/io/camunda/testing/utils/RecordStreamManager.java
pihme: Also not sure whether the class is helpful, to be honest. Now BPMNAssert and this class are sort of tied together, but the coupling is less obvious. This is not to be understood as a veto though. If you have ideas on how this class will develop going forward, I am happy to discuss these. But right now it looks like you separated concerns, but not the underlying coupling.

src/main/java/io/camunda/testing/utils/ProcessEventInspections.java
```diff
@@ -0,0 +1,59 @@
...
+  public ProcessInstanceAssert assertThatFirstProcessInstance() {
+    return assertThatProcessInstance(0);
+  }
+
+  public ProcessInstanceAssert assertThatLastProcessInstance() {
+    final List<Long> processInstanceKeys = getProcessInstanceKeys();
+    return assertThatProcessInstance(processInstanceKeys, processInstanceKeys.size() - 1);
+  }
+
+  public ProcessInstanceAssert assertThatProcessInstance(final int times) {
```
pihme, line 38: :x: rename `times`-> `index` 

Also please add JavaDocs to these methods. This in particular is not self-explanatory, in my mind.

pihme (changes requested):
Lots of good stuff here. However, I would like to propose a change to the separation of concerns between `InspectionUtility` and assertions. (see my first comment)

src/main/java/io/camunda/testing/utils/ProcessEventInspections.java
```diff
@@ -17,39 +15,83 @@ public ProcessEventInspections(final ProcessEventRecordStreamFilter filter) {
     this.filter = filter;
   }
 
+  /**
+   * Filters the process instances to only include instances that were triggered by a given timer id
+   *
+   * @param timerId The id of the timer
```
pihme, line 21: :wrench: Rephrase suggestion: "The element ID of the timer event"

pihme (approved):
Love it!

PR #65, opened 2021-11-16
Refactor ZeebeAssertionsExtension

### LINKED ISSUE #62: Add error handling to the ZeebeAssertionsExtension
**Description**

In the `ZeebeAssertionsExtensions` we use reflection to find specific fields in tests. At the moment we assume these fields are always available. If they are not the `Optional.get()` function will throw default exceptions.

We should handle these default exceptions and provide detailed feedback about what went wrong (eg. what fields we are missing in the test class).

remcowesterhoud:
## Description

- Split code into separate methods
- Add error handling
- Add unit tests

## Related issues

closes #62

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/main/java/io/camunda/testing/extensions/ZeebeAssertionsExtension.java
+++ b/src/main/java/io/camunda/testing/extensions/ZeebeAssertionsExtension.java
@@ -17,9 +17,8 @@ public class ZeebeAssertionsExtension implements BeforeEachCallback, AfterEachCa
 
   @Override
   public void beforeEach(final ExtensionContext extensionContext) throws Exception {
-    final Object testInstance = extensionContext.getRequiredTestInstance();
-    storeRecordStreamSource(extensionContext, testInstance);
-    storeZeebeClient(extensionContext, testInstance);
+    storeRecordStreamSource(extensionContext);
+    storeZeebeClient(extensionContext);
   }
 
   @Override
@@ -31,13 +30,12 @@ public class ZeebeAssertionsExtension implements BeforeEachCallback, AfterEachCa
     zeebeClient.close();
   }
 
-  private void storeRecordStreamSource(
-      final ExtensionContext extensionContext, final Object testInstance)
+  private void storeRecordStreamSource(final ExtensionContext extensionContext)
       throws IllegalAccessException {
     final Field recordStreamField = getRecordStreamField(extensionContext);
     ReflectionUtils.makeAccessible(recordStreamField);
     final RecordStreamSource recordStreamSource =
-        (RecordStreamSource) recordStreamField.get(testInstance);
+        (RecordStreamSource) recordStreamField.get(extensionContext.getRequiredTestInstance());
     RecordStreamSourceStore.init(recordStreamSource);
   }
 
@@ -50,14 +48,16 @@ public class ZeebeAssertionsExtension implements BeforeEachCallback, AfterEachCa
         () ->
             new IllegalStateException(
                 "Expected a field of type RecordStreamSource to be declared in the test class, "
-                    + "but none has been found"));
+                    + "but none has been found. Please make sure a field of type RecordStreamSource"
+                    + " has been declared in the test class."));
   }
 
-  private void storeZeebeClient(final ExtensionContext extensionContext, final Object testInstance)
+  private void storeZeebeClient(final ExtensionContext extensionContext)
       throws IllegalAccessException {
     final Field zeebeClientField = getZeebeClientField(extensionContext);
     ReflectionUtils.makeAccessible(zeebeClientField);
-    final ZeebeClient zeebeClient = (ZeebeClient) zeebeClientField.get(testInstance);
+    final ZeebeClient zeebeClient =
+        (ZeebeClient) zeebeClientField.get(extensionContext.getRequiredTestInstance());
     getStore(extensionContext).put(KEY_ZEEBE_CLIENT, zeebeClient);
   }
 
@@ -70,7 +70,8 @@ public class ZeebeAssertionsExtension implements BeforeEachCallback, AfterEachCa
         () ->
             new IllegalStateException(
                 "Expected a field of type ZeebeClient to be declared in the test class, "
-                    + "but none has been found"));
+                    + "but none has been found. Please make sure a field of type ZeebeClient "
+                    + "has been declared in the test class."));
   }
 
   private ExtensionContext.Store getStore(final ExtensionContext context) {
--- a/src/main/java/io/camunda/testing/extensions/ZeebeAssertionsExtension.java
+++ b/src/main/java/io/camunda/testing/extensions/ZeebeAssertionsExtension.java
@@ -22,7 +22,11 @@ public class ZeebeAssertionsExtension implements BeforeEachCallback, AfterEachCa
         Arrays.stream(extensionContext.getTestClass().get().getDeclaredFields())
             .filter(field -> field.getType() == RecordStreamSource.class)
             .findFirst();
-    final Field recordStreamField = recordStreamOptional.get();
+    final Field recordStreamField = recordStreamOptional.orElseThrow(
+        () ->
+            new IllegalStateException(
+                "No RecordStreamSource has been found. Please make sure a "
+                    + "RecordStreamSource field has been declared in the test class."));
     ReflectionUtils.makeAccessible(recordStreamField);
     final Object testInstance = extensionContext.getTestInstance().get();
     final RecordStreamSource recordStreamSource =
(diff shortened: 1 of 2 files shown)


remcowesterhoud:
@korthout Would you be willing to review this PR since Peter is on holiday? I'm happy to answer any questions you might have.

korthout:
Hi @remcowesterhoud. 👋  

Sure I'd love to help, but I'm having some trouble understanding these changes.

Currently there is one big commit that is called `refactor: Refactor ...`, but it doesn't really help me understand things. I have the following questions that I want to find answers to:
- Why do these changes resolve #62? Everything seems to indicate that this is a refactoring, and does not improve the error output.
- There are tests added, do these verify that the behavior is still the same after the refactoring?
- Did this test also pass before the refactorings?

To make my life easier, can you please split up the changes into smaller commits?

remcowesterhoud:
Thanks @korthout. 

I've split the commits into 2 separate commits. One with the added error handling, and one with the refactor. Hope this helps!

To answer your questions:

**Why do these changes resolve Add error handling to the ZeebeAssertionsExtension #62? Everything seems to indicate that this is a refactoring, and does not improve the error output.**
Before we would use `get()` the `recordStreamOptional` and the `zeebeClientOptional`. We did not check whether or not these fields were present. With this change we will make sure they are present, or else throw an exception explaining that these fields are required.

**There are tests added, do these verify that the behavior is still the same after the refactoring?**
The error handling is new as explained in in the question above. Therefore, the unit tests are also new.

**Did this test also pass before the refactorings?**
See above answers 😄

korthout:
@remcowesterhoud Thanks this helps a lot! 👍

src/test/java/io/camunda/testing/extensions/ZeebeAssertionsExtensionTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    public void testExtensionThrowsExceptionWhenMissingRecordStreamSource() {
+      // given
+      final ZeebeAssertionsExtension extension = new ZeebeAssertionsExtension();
+      final ExtensionContext extensionContext = mock(ExtensionContext.class);
+      Mockito.when(extensionContext.getTestClass()).thenReturn(Optional.of(this.getClass()));
+      Mockito.when(extensionContext.getTestInstance()).thenReturn(Optional.of(this));
+
+      // when & then
```
korthout, line 28: 🙃  this `//when` comment belongs above the `Mockito.when` calls

src/test/java/io/camunda/testing/extensions/ZeebeAssertionsExtensionTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.mockito.Mockito;
+
+class ZeebeAssertionsExtensionTest {
+
+  @Nested
+  class MissingRecordStreamSource {
+
+    private ZeebeClient client;
```
korthout, line 18: 🔧 Maybe add a comment to this field, explaining why it is necessary for this test case, i.e. in order for `RecordStreamSource` to be the only one that is missing it needs to have all other required fields, incl `ZeebeClient`.

src/test/java/io/camunda/testing/extensions/ZeebeAssertionsExtensionTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import io.camunda.zeebe.client.ZeebeClient;
+import java.util.Optional;
+import org.camunda.community.eze.RecordStreamSource;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.mockito.Mockito;
+
+class ZeebeAssertionsExtensionTest {
```
korthout, line 13: 👍  I really like the simplicity of these tests, they are easy to understand. 🏅

src/main/java/io/camunda/testing/extensions/ZeebeAssertionsExtension.java
```diff
@@ -11,50 +11,66 @@
...
+    ReflectionUtils.makeAccessible(recordStreamField);
+    final RecordStreamSource recordStreamSource =
+        (RecordStreamSource) recordStreamField.get(testInstance);
+    RecordStreamSourceStore.init(recordStreamSource);
+  }
+
+  private Field getRecordStreamField(final ExtensionContext extensionContext) {
     final Optional<Field> recordStreamOptional =
-        Arrays.stream(extensionContext.getTestClass().get().getDeclaredFields())
+        Arrays.stream(extensionContext.getRequiredTestClass().getDeclaredFields())
```
korthout, line 46: 🙃  This change it not a refactoring, i.e. it changes the error message if the test class is missing for the extensionContext. TBH, I'm not sure whether that can actually happen, which brings me to something actually interesting.

🔧  There is something to say for dealing with the case that the class or the instance is missing (`getTestClass().orElseThrow(...)`), in which case we could help the user with a helpful error message similar to the missing required fields. All this depends on whether the class or the instance (below) can ever be missing. If they can never be missing than I prefer `getRequiredTestClass()`.

src/main/java/io/camunda/testing/extensions/ZeebeAssertionsExtension.java
```diff
@@ -11,50 +11,66 @@
...
 import org.junit.platform.commons.util.ReflectionUtils;
 
-// TODO rewrite this mess
 public class ZeebeAssertionsExtension implements BeforeEachCallback, AfterEachCallback {
 
   private static final String KEY_ZEEBE_CLIENT = "ZEEBE_CLIENT";
 
   @Override
   public void beforeEach(final ExtensionContext extensionContext) throws Exception {
+    final Object testInstance = extensionContext.getRequiredTestInstance();
```
korthout, line 20: 🙃  For understandability, I'd move this line into both of the store methods, duplicated just like the `getRequiredTestClass()` calls. That way the beforeEach becomes more descriptive, first do this, then do that.

korthout (changes requested):
👍  Thanks for splitting up the changes and your explanations, it made reviewing much more easy. 

I found some things important to raise and the rest are suggestions that are at a take-it-or-leave-it level. Please have a look.

Additionally, please update the pull request summary to reflect the actual changes.

src/test/java/io/camunda/testing/extensions/ZeebeAssertionsExtensionTest.java
```diff
@@ -0,0 +1,59 @@
...
+  @Nested
+  class MissingRecordStreamSource {
+
+    private ZeebeClient client;
+
+    @Test
+    public void testExtensionThrowsExceptionWhenMissingRecordStreamSource() {
+      // given
+      final ZeebeAssertionsExtension extension = new ZeebeAssertionsExtension();
+      final ExtensionContext extensionContext = mock(ExtensionContext.class);
```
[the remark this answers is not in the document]
remcowesterhoud: Ah I should've specified that this project uses a Java 8 language level. That way users with that are still running on older java version are able to use it as well 😄

src/main/java/io/camunda/testing/extensions/ZeebeAssertionsExtension.java
```diff
@@ -22,7 +22,11 @@ public void beforeEach(final ExtensionContext extensionContext) throws Exception
         Arrays.stream(extensionContext.getTestClass().get().getDeclaredFields())
             .filter(field -> field.getType() == RecordStreamSource.class)
             .findFirst();
-    final Field recordStreamField = recordStreamOptional.get();
+    final Field recordStreamField = recordStreamOptional.orElseThrow(
+        () ->
+            new IllegalStateException(
+                "No RecordStreamSource has been found. Please make sure a "
+                    + "RecordStreamSource field has been declared in the test class."));
```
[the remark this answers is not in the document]
remcowesterhoud: I was not aware about these guidelines. I have applied them to the error messages in this PR.

As for the other messages in this project, they are all assertions. We have decided on a way to formulate the assertion messages so we should keep these the way they are for now. If user feedback suggests that these message are unclear and need to be improved we can do this at a later stage.

src/main/java/io/camunda/testing/extensions/ZeebeAssertionsExtension.java
remcowesterhoud: I thought about this. I don't know in which scenario the test class could not be present. Maybe when using the extension on a non-testclass. But when that happens the `beforeEach()` wouldn't be executed in the first place.

I will stick with `getRequiredTestClass()` for now. If we ever run into this problem we can find a solution at that time.

remcowesterhoud:
Thanks for your review @korthout! 🍪

I've commented / applied your suggestion.

src/main/java/io/camunda/testing/extensions/ZeebeAssertionsExtension.java
```diff
@@ -49,8 +49,8 @@ private Field getRecordStreamField(final ExtensionContext extensionContext) {
     return recordStreamOptional.orElseThrow(
         () ->
             new IllegalStateException(
-                "No RecordStreamSource has been found. Please make sure a "
-                    + "RecordStreamSource field has been declared in the test class."));
+                "Expected a field of type RecordStreamSource to be declared in the test class, "
+                    + "but none has been found"));
```
korthout, line 51: 🔧 To be honest, now I miss the user hint. Would be nice to add it behind it.

Something like:
```
"Expected a field of type RecordStreamSource to be declared in the test class, "
    + "but none has been found. Please make sure a field of type RecordStreamSource "
    + "has been declared in the test class."
```

korthout (approved):
LGTM 🥨 

Please consider my last comment before merging

PR #66, opened 2021-11-19
feat: Deploy snapshot version to artifactory

### LINKED ISSUE #25: Deploy to artifactory
**Description**

To make it easy for users to use this testing library in their own project we want to publish it to artifactory. To achieve this we should expand our pipeline with a deploy step that gets triggered when a new release is created.

remcowesterhoud:
## Description

- Add import secrets step in pipeline to get the secrets out of Hashicorp vault
- Snapshots can be published by triggering the GitHub action manually.
- Use the community-action-maven-release GitHub action to deploy a snapshot version to artifactory. This action also makes it possible to release to maven central, for now the maven credentials have been filled with dummy values. Nothing should be pushed to maven central yet. **This will cause the pipeline to fail when you want to deploy a SNAPSHOT**, however the SNAPSHOT will still publish to Artifactory. It might be somewhat confusing, but I figured it would be fine to get an initial SNAPSHOT out there so we could gather feedback.
- Use the community-hub-release-parent pom. This is required for the community release action. To make sure the artifacts land in the correct repository we needed to override some properties from the parent pom.

## Related issues

closes #25 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2,12 +2,18 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
+  <parent>
+    <groupId>org.camunda.community</groupId>
+    <artifactId>community-hub-release-parent</artifactId>
+    <version>1.2.2</version>
+  </parent>
+
   <groupId>io.camunda</groupId>
-  <artifactId>camunda-cloud-testing</artifactId>
+  <artifactId>zeebe-bpmn-assert</artifactId>
   <version>0.0.1-SNAPSHOT</version>
   <packaging>jar</packaging>
 
-  <name>Camunda Cloud Testing</name>
+  <name>Zeebe BPMN Assert</name>
 
   <scm>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -2,12 +2,18 @@
...
 
+  <parent>
+    <groupId>org.camunda.community</groupId>
+    <artifactId>community-hub-release-parent</artifactId>
+    <version>1.2.2</version>
+  </parent>
+
   <groupId>io.camunda</groupId>
-  <artifactId>camunda-cloud-testing</artifactId>
+  <artifactId>zeebe-bpmn-assert</artifactId>
```
pihme, line 12: Do we want to reflect the name change also in the package structure?

pihme (approved):
Looks good!

One thing I am wondering, though, is whether we need to change the settings for this repo.

The Vault GitHub action log states that the variables will be obfuscated in logs and such, but if anyone could modify the workflow, they could print out the variables in another way. And our settings right now is that external collaborators only need to be authenticated once.

But maybe this is not needed anyway, because an outside collaborator cannot access our secreted in vault in the first place. I am really not sure how this works in detail.

pom.xml
remcowesterhoud: Yes, we will do this in a different PR as discussed

remcowesterhoud:
The vault is accessed by the secrets we have configured in this GitHub repo. External collaborators would only by able to print these secrets if we somehow missed this during a code review.

PR #69, opened 2021-11-23
Inspect Call Activity

### LINKED ISSUE #46: Add means to assert on process instances that were created indirectly

remcowesterhoud:
## Description

When a process instance is started by a Call Activity element, there is no entry point for the assertions. This makes it impossible to test these process instances. With this change I've introduces a `ProcessInstanceInspections` class which will make it possible to find these process instances and assert over them.

## Related issues

closes #46 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/main/java/io/camunda/testing/utils/ProcessInstanceInspections.java
+++ b/src/main/java/io/camunda/testing/utils/ProcessInstanceInspections.java
@@ -1,0 +1,101 @@
+package io.camunda.testing.utils;
+
+import io.camunda.testing.filters.ProcessInstanceRecordStreamFilter;
+import io.camunda.testing.utils.model.InspectedProcessInstance;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class ProcessInstanceInspections {
+
+  private final ProcessInstanceRecordStreamFilter filter;
+
+  public ProcessInstanceInspections(final ProcessInstanceRecordStreamFilter filter) {
+    this.filter = filter;
+  }
+
+
+  /**
+   * Filters the process instances to only include instances that were started by the given parent
+   * process instance key
+   *
+   * @param key The key of the parent process instance
+   * @return this {@link ProcessInstanceInspections}
+   */
+  public ProcessInstanceInspections startedByProcessInstance(final long key) {
+    return new ProcessInstanceInspections(filter.withParentProcessInstanceKey(key));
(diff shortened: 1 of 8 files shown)


src/main/java/io/camunda/testing/utils/ProcessInstanceInspections.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Filters the process instances to only include instances that were started by the given parent
+   * process instance key
+   *
+   * @param key The key of the parent process instance
+   * @return this {@link ProcessInstanceInspections}
+   */
+  public ProcessInstanceInspections startedByProcessInstance(final long key) {
```
pihme, line 24: :x: rename to `withParentProcessInstanceKey` for consistency with assert and also better following of naming conventions here.

src/main/java/io/camunda/testing/utils/ProcessInstanceInspections.java
```diff
@@ -0,0 +1,100 @@
+package io.camunda.testing.utils;
+
+import io.camunda.testing.filters.ProcessInstanceRecordStreamFilter;
+import io.camunda.testing.utils.model.InspectedProcessInstance;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class ProcessInstanceInspections {
```
pihme, line 9: :x: Please also add an example to the Readme like you did for "started by Timer"

(I like the new readme, by the way :trophy:)

pihme (approved):
One change and one addition requested.

If you want, I can look at it again. But you can also merge it right away, if you want.

PR #70, opened 2021-11-23
Rename packages

remcowesterhoud:
## Description

Since that artifact id of the project has been changed to `zeebe-bpmn-assert` we should change the package structure to match this name.

We should also rename the `utils` package as it is not descriptive enough. It only contains inspections and the `RecordStreamSourceStore` at this moment. We can rename the package to `inspections` and move the `RecordStreamSourceStore` out of it.

## Definition of Done

Code changes:

Testing:

Documentation:


remcowesterhoud:
@pihme I've renamed the packages. These names are my proposal, please let me know if you had something else in mind.

`io.camunda.testing` > `io.camunda.zeebe.bpmnassert`
`utils` > `inspections`

I had to go with `bpmnassert` as one because it's not allowed to use keywords in package names.

pihme (approved):
I like it!

PR #76, opened 2021-11-26
Implement test engine

### LINKED ISSUE #24: Copy EZE to this project
**Description**

For the first MVP our testing solution will depend on [EZE](https://github.com/camunda-community-hub/eze). We do not want to depend on a community project. Therefore, we need to create a copy of EZE and add it to this project.

Things to note:
- EZE is currently written in Kotlin. We want our copy to be written in Java
- We want our version of EZE to use Java 17

remcowesterhoud:
## Description

EZE has been copied and translated into Java.
Still to do to entirely get rid of the EZE dependency:
- Modify `ZeebeAssertionsExtension` so that it will take over the functionality of the `@EmbeddedZeebeEngine`
- Modify tests to use our own classes
- Add our own dependency on Zeebe

## Related issues

closes #24 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/main/java/io/camunda/zeebe/bpmnassert/testengine/IdleStateMonitor.java
+++ b/src/main/java/io/camunda/zeebe/bpmnassert/testengine/IdleStateMonitor.java
@@ -1,0 +1,73 @@
+package io.camunda.zeebe.bpmnassert.testengine;
+
+import io.camunda.zeebe.engine.processing.streamprocessor.StreamProcessorListener;
+import io.camunda.zeebe.engine.processing.streamprocessor.TypedRecord;
+import io.camunda.zeebe.logstreams.log.LogStreamReader;
+import io.camunda.zeebe.logstreams.log.LoggedEvent;
+import io.camunda.zeebe.logstreams.storage.LogStorage;
+import java.util.ArrayList;
+import java.util.List;
+
+final class IdleStateMonitor implements LogStorage.CommitListener, StreamProcessorListener {
+
+  private final List<Runnable> callbacks = new ArrayList<>();
+  private final LogStreamReader reader;
+  private long lastEventPosition = -1L;
+  private long lastProcessedPosition = -1;
+
+  IdleStateMonitor(final InMemoryLogStorage logStorage, final LogStreamReader logStreamReader) {
+    logStorage.addCommitListener(this);
+
+    reader = logStreamReader;
+  }
+
+  public void addCallback(final Runnable callback) {
+    synchronized (callbacks) {
+      callbacks.add(callback);
+    }
+    checkIdleState();
+  }
+
+  private void checkIdleState() {
+    if (isInIdleState()) {
+      synchronized (callbacks) {
+        callbacks.forEach(Runnable::run);
+        callbacks.clear();
+      }
+    }
+  }
+
+  private boolean isInIdleState() {
+    forwardToLastEvent();
+    return lastEventPosition == lastProcessedPosition;
+  }
+
+  private void forwardToLastEvent() {
+    while (reader.hasNext()) {
+      lastEventPosition = reader.next().getPosition();
+    }
+  }
+
+  @Override
+  public void onCommit() {
+    checkIdleState();
+  }
+
+  @Override
+  public void onProcessed(final TypedRecord<?> processedCommand) {
+    lastProcessedPosition = processedCommand.getPosition();
+    checkIdleState();
+  }
+
+  @Override
+  public void onSkipped(final LoggedEvent skippedRecord) {
+    lastProcessedPosition = skippedRecord.getPosition();
+    checkIdleState();
+  }
+
+  @Override
+  public void onReplayed(final long lastReplayedEventPosition, final long lastReadRecordPosition) {
+    lastProcessedPosition = lastReplayedEventPosition;
+    checkIdleState();
+  }
+}
--- a/src/main/java/io/camunda/zeebe/bpmnassert/testengine/PartitionCommandSenderImpl.java
+++ b/src/main/java/io/camunda/zeebe/bpmnassert/testengine/PartitionCommandSenderImpl.java
@@ -1,0 +1,34 @@
+package io.camunda.zeebe.bpmnassert.testengine;
+
+import io.camunda.zeebe.engine.processing.message.command.PartitionCommandSender;
+import io.camunda.zeebe.engine.processing.message.command.SubscriptionCommandMessageHandler;
+import io.camunda.zeebe.util.buffer.BufferWriter;
+import org.agrona.concurrent.UnsafeBuffer;
+
+public class PartitionCommandSenderImpl implements PartitionCommandSender {
+
+  private final SubscriptionCommandMessageHandler subscriptionHandler;
+  private final int partitionId;
+
+  public PartitionCommandSenderImpl(
+      final SubscriptionCommandMessageHandler subscriptionHandler, final int partitionId) {
+    this.subscriptionHandler = subscriptionHandler;
+    this.partitionId = partitionId;
+  }
+
+  @Override
+  public boolean sendCommand(final int receiverPartitionId, final BufferWriter command) {
+    if (receiverPartitionId != partitionId) {
+      throw new RuntimeException(
+          String.format(
+              "Expected receiverPartitionId to be %d, but was %d",
+              partitionId, receiverPartitionId));
+    }
+
+    final byte[] bytes = new byte[command.getLength()];
+    final UnsafeBuffer commandBuffer = new UnsafeBuffer(bytes);
+    command.write(commandBuffer, 0);
+    subscriptionHandler.apply(bytes);
+    return true;
+  }
+}
--- a/src/main/java/io/camunda/zeebe/bpmnassert/testengine/RecordStreamSource.java
+++ b/src/main/java/io/camunda/zeebe/bpmnassert/testengine/RecordStreamSource.java
@@ -1,0 +1,38 @@
+package io.camunda.zeebe.bpmnassert.testengine;
+
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.value.*;
+import io.camunda.zeebe.protocol.record.value.deployment.Process;
+
+public interface RecordStreamSource {
+
+  Iterable<Record<?>> records();
+
+  Iterable<Record<ProcessInstanceRecordValue>> processInstanceRecords();
+
+  Iterable<Record<JobRecordValue>> jobRecords();
+
+  Iterable<Record<JobBatchRecordValue>> jobBatchRecords();
+
+  Iterable<Record<DeploymentRecordValue>> deploymentRecords();
+
+  Iterable<Record<Process>> processRecords();
+
+  Iterable<Record<VariableRecordValue>> variableRecords();
+
+  Iterable<Record<VariableDocumentRecordValue>> variableDocumentRecords();
+
+  Iterable<Record<IncidentRecordValue>> incidentRecords();
+
+  Iterable<Record<TimerRecordValue>> timerRecords();
+
+  Iterable<Record<MessageRecordValue>> messageRecords();
+
+  Iterable<Record<MessageSubscriptionRecordValue>> messageSubscriptionRecords();
+
+  Iterable<Record<MessageStartEventSubscriptionRecordValue>> messageStartEventSubscriptionRecords();
+
+  Iterable<Record<ProcessMessageSubscriptionRecordValue>> processMessageSubscriptionRecords();
+
+  void print(final boolean compact);
+}
--- a/src/main/java/io/camunda/zeebe/bpmnassert/testengine/RecordStreamSourceImpl.java
+++ b/src/main/java/io/camunda/zeebe/bpmnassert/testengine/RecordStreamSourceImpl.java
@@ -1,0 +1,180 @@
+package io.camunda.zeebe.bpmnassert.testengine;
+
+import io.camunda.zeebe.engine.processing.streamprocessor.TypedEventRegistry;
+import io.camunda.zeebe.logstreams.log.LogStreamReader;
+import io.camunda.zeebe.logstreams.log.LoggedEvent;
+import io.camunda.zeebe.protocol.impl.record.CopiedRecord;
+import io.camunda.zeebe.protocol.impl.record.RecordMetadata;
+import io.camunda.zeebe.protocol.impl.record.UnifiedRecordValue;
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.RecordValue;
+import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.protocol.record.value.DeploymentRecordValue;
+import io.camunda.zeebe.protocol.record.value.IncidentRecordValue;
+import io.camunda.zeebe.protocol.record.value.JobBatchRecordValue;
+import io.camunda.zeebe.protocol.record.value.JobRecordValue;
+import io.camunda.zeebe.protocol.record.value.MessageRecordValue;
+import io.camunda.zeebe.protocol.record.value.MessageStartEventSubscriptionRecordValue;
+import io.camunda.zeebe.protocol.record.value.MessageSubscriptionRecordValue;
+import io.camunda.zeebe.protocol.record.value.ProcessInstanceRecordValue;
+import io.camunda.zeebe.protocol.record.value.ProcessMessageSubscriptionRecordValue;
+import io.camunda.zeebe.protocol.record.value.TimerRecordValue;
+import io.camunda.zeebe.protocol.record.value.VariableDocumentRecordValue;
+import io.camunda.zeebe.protocol.record.value.VariableRecordValue;
+import io.camunda.zeebe.protocol.record.value.deployment.Process;
+import io.camunda.zeebe.test.util.record.CompactRecordLogger;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+public class RecordStreamSourceImpl implements RecordStreamSource {
+
+  private final LogStreamReader logStreamReader;
+  private final int partitionId;
+  private final List<Record<?>> records = new ArrayList<>();
+  private long lastPosition = -1L;
+
+  public RecordStreamSourceImpl(final LogStreamReader logStreamReader, final int partitionId) {
+    this.logStreamReader = logStreamReader;
+    this.partitionId = partitionId;
+  }
+
+  @Override
+  public Iterable<Record<?>> records() {
+    updateWithNewRecords();
+    return Collections.unmodifiableList(records);
+  }
+
+  <T extends RecordValue> Iterable<Record<T>> recordsOfValueType(final ValueType valueType) {
+    final Stream<Record<T>> stream =
+        (Stream)
+            StreamSupport.stream(records().spliterator(), false)
+                .filter((record) -> record.getValueType() == valueType);
+    return stream::iterator;
+  }
+
+  @Override
+  public Iterable<Record<ProcessInstanceRecordValue>> processInstanceRecords() {
+    return recordsOfValueType(ValueType.PROCESS_INSTANCE);
+  }
+
+  @Override
+  public Iterable<Record<JobRecordValue>> jobRecords() {
+    return recordsOfValueType(ValueType.JOB);
+  }
+
+  @Override
+  public Iterable<Record<JobBatchRecordValue>> jobBatchRecords() {
+    return recordsOfValueType(ValueType.JOB_BATCH);
+  }
+
+  @Override
+  public Iterable<Record<DeploymentRecordValue>> deploymentRecords() {
+    return recordsOfValueType(ValueType.DEPLOYMENT);
+  }
+
+  @Override
+  public Iterable<Record<Process>> processRecords() {
+    return recordsOfValueType(ValueType.PROCESS);
+  }
+
+  @Override
+  public Iterable<Record<VariableRecordValue>> variableRecords() {
+    return recordsOfValueType(ValueType.VARIABLE);
+  }
+
+  @Override
+  public Iterable<Record<VariableDocumentRecordValue>> variableDocumentRecords() {
+    return recordsOfValueType(ValueType.VARIABLE_DOCUMENT);
+  }
+
+  @Override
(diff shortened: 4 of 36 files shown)


src/main/java/io/camunda/zeebe/bpmnassert/testengine/PartitionCommandSenderImpl.java
```diff
@@ -0,0 +1,35 @@
...
+import io.camunda.zeebe.engine.processing.message.command.SubscriptionCommandMessageHandler;
+import io.camunda.zeebe.util.buffer.BufferWriter;
+import java.util.Map;
+import org.agrona.concurrent.UnsafeBuffer;
+
+public class PartitionCommandSenderImpl implements PartitionCommandSender {
+
+  private final Map<Integer, SubscriptionCommandMessageHandler> subscriptionHandlers;
+
+  public PartitionCommandSenderImpl(
```
pihme, line 13: :wrench: Refactoring proposal

We know in advance that we have only one partition. So we could pass in the handler directly and simplify the call sites leading up to here.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/RecordStreamSource.java
```diff
@@ -0,0 +1,93 @@
...
+      messageStartEventSubscriptionRecords() {
+    return recordsOfValueType(ValueType.MESSAGE_START_EVENT_SUBSCRIPTION);
+  }
+
+  default Iterable<Record<ProcessMessageSubscriptionRecordValue>>
+      processMessageSubscriptionRecords() {
+    return recordsOfValueType(ValueType.PROCESS_MESSAGE_SUBSCRIPTION);
+  }
+
+  default void print(final boolean compact) {
```
pihme, line 81: :wrench: Maybe add `toString()`method that calls print

src/main/java/io/camunda/zeebe/bpmnassert/testengine/SinglePartitionDeploymentDistributor.java
```diff
@@ -0,0 +1,15 @@
+package io.camunda.zeebe.bpmnassert.testengine;
+
+import io.camunda.zeebe.engine.processing.deployment.distribute.DeploymentDistributor;
+import io.camunda.zeebe.util.sched.future.ActorFuture;
+import io.camunda.zeebe.util.sched.future.CompletableActorFuture;
+import org.agrona.DirectBuffer;
+
+public class SinglePartitionDeploymentDistributor implements DeploymentDistributor {
```
pihme, line 8: :thought_balloon: We can also align on whether to use final for classes and whether or not to limit visibility to the minimum. In the DB folder must stuff is final and package private. On the other hand, it's also frustrating when you want to extend this library and you cannot because we put all the final keywords in and hid stuff.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/SubscriptionCommandSenderFactory.java
```diff
@@ -0,0 +1,40 @@
...
+import io.camunda.zeebe.engine.processing.message.command.SubscriptionCommandSender;
+import io.camunda.zeebe.logstreams.log.LogStreamRecordWriter;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+public class SubscriptionCommandSenderFactory {
+
+  final Map<Integer, LogStreamRecordWriter> streamWritersByPartition;
```
pihme, line 13: see other comment on we know it's a single partition already

src/main/java/io/camunda/zeebe/bpmnassert/testengine/ZeebeEngine.java
```diff
@@ -0,0 +1,20 @@
+package io.camunda.zeebe.bpmnassert.testengine;
+
+import io.camunda.zeebe.client.ZeebeClient;
+
+public interface ZeebeEngine extends RecordStreamSource {
+
+  void start();
+
+  void stop();
```
pihme, line 9: :thought_balloon: One thing I pondered is whether we want to add a `reset()` method in the future. This method could clear the in memory database and in memory log storage.

Especially when we think about running this in a test container, we would gain test execution speed if after each test we just do a reset instead of restarting the whole test container.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/ZeebeEngine.java
```diff
@@ -0,0 +1,20 @@
...
+
+  ZeebeClient createClient();
+
+  String getGatewayAddress();
+
+  ZeebeEngineClock clock();
```
pihme, line 15: :wrench: refactoring proposal. Instead of returning the actor clock, we could just offer the method `increaseTime()`here. This way we could get rid of `ZeebeEngineClock` and `ZeebeEngineClockImpl`

src/main/java/io/camunda/zeebe/bpmnassert/testengine/ZeebeEngineImpl.java
```diff
@@ -0,0 +1,75 @@
...
+
+import io.camunda.zeebe.client.ZeebeClient;
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.util.sched.clock.ControlledActorClock;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Supplier;
+
+public class ZeebeEngineImpl implements ZeebeEngine {
+
+  public static final int PORT = 26500;
```
pihme, line 11: Please make dynmic

src/test/java/io/camunda/zeebe/bpmnassert/testengine/PrintRecordStreamExtension.java
```diff
@@ -0,0 +1,31 @@
...
+ * except in compliance with the Zeebe Community License 1.1.
+ */
+package io.camunda.zeebe.bpmnassert.testengine;
+
+import java.lang.reflect.Field;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.TestWatcher;
+import org.junit.platform.commons.util.ReflectionUtils;
+
+class PrintRecordStreamExtension implements TestWatcher {
```
pihme, line 15: :question: Why is this in the test folder? I would expect it in the source folder

pihme (changes requested):
Great work!

Several comments and change requests.

src/test/java/io/camunda/zeebe/bpmnassert/testengine/EngineClientTest.java
```diff
@@ -0,0 +1,799 @@
...
+      // when - then
+      zeebeClient.newFailCommand(job.getKey())
+          .retries(0)
+          .errorMessage("This failed oops.")
+          .send()
+          .join();
+
+      await().untilAsserted(() -> {
+        final Optional<Record<JobRecordValue>> failedJob = new JobRecordStreamFilter(
+            zeebeEngine.jobRecords())
```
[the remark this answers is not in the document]
pihme: Even worse offender, in my mind:

```
      System.out.println("===== Test failed! Printing records from the stream:");
      zeebeEngine.print(true);
```

src/test/java/io/camunda/zeebe/bpmnassert/testengine/db/InMemoryZeebeDbTransactionTest.java
```diff
@@ -0,0 +1,479 @@
...
+          // when
+          final DbLong zbLong = oneColumnFamily.get(oneKey);
+
+          // then
+          assertThat(zbLong).isNull();
+        });
+  }
+
+  @Test
+  public void shouldStartNewTransaction() throws Exception {
```
remcowesterhoud, line 105: 🔧 This test doesn't fit the name. It puts 3 values in the database, commits them and check that they exist. It never checks if the transaction was completed and a new one has been started. The transaction might still be active, because if the keys are available in the transaction cache `exists()` will also return true.

src/test/java/io/camunda/zeebe/bpmnassert/testengine/db/InMemoryZeebeDbTransactionTest.java
```diff
@@ -0,0 +1,479 @@
...
+          final ZeebeDbTransaction sameTransaction = transactionContext.getCurrentTransaction();
+          sameTransaction.run(
+              () -> {
+                oneColumnFamily.put(oneKey, oneValue);
+                twoColumnFamily.put(twoKey, twoValue);
+                threeColumnFamily.put(threeKey, threeValue);
+              });
+          sameTransaction.commit();
+
+          // then it is committed but available in this transaction
```
remcowesterhoud, line 204: ❓ I'm not sure about this test. I feel like this makes it seem like `sameTransaction` is still active. However, this is not the case. With the `commit()` this transaction was ended. 2 lines down from this one a new transaction gets started with `oneColumnFamily.delete(oneKey)`.

I don't know what the intended behaviour of this test is. The test says it should not reopen a transaction, but a new transaction does get started. So I'm a little bit confused here.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/db/InMemoryZeebeDbTransactionContext.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  private void runInNewTransaction(final TransactionOperation operations) throws Exception {
+    try {
+      transaction.resetTransaction();
+      operations.run();
+      transaction.commit();
+    } finally {
+      transaction.rollback();
```
remcowesterhoud, line 35: ❌ We should only rollback the transaction in case of an exception. If we've successfully committed there shouldn't be a need for this.

src/test/java/io/camunda/zeebe/bpmnassert/testengine/db/InMemoryZeebeDbTransactionTest.java
```diff
@@ -0,0 +1,479 @@
...
+    transactionContext.runInTransaction(
+        () -> {
+          oneColumnFamily.put(oneKey, oneValue);
+          final DbLong value = oneColumnFamily.get(oneKey);
+          actualValue.set(value.getValue());
+        });
+
+    // then
+    assertThat(actualValue.get()).isEqualTo(-1);
+    assertThat(oneColumnFamily.get(oneKey).getValue()).isEqualTo(-1);
```
remcowesterhoud, line 264: 🙃 I don't think this assertion should be part of this test. The test wants to assert that we can get values in a transaction. At this stage the transaction has been completed, so getting this value has nothing to do with getting a value in a transaction. This case is already being verified in other tests.

src/test/java/io/camunda/zeebe/bpmnassert/testengine/db/InMemoryZeebeDbTransactionTest.java
```diff
@@ -0,0 +1,479 @@
...
+
+import io.camunda.zeebe.db.*;
+import io.camunda.zeebe.db.impl.DbLong;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public final class InMemoryZeebeDbTransactionTest {
```
remcowesterhoud, line 20: 🔧 An interesting case (which I believe will fail) would be to start a transaction. In the transaction a key gets deleted and the same key get re-added:
```java
columnFamily.delete(key);
columnFamily.put(key, value);
```
In this case I'd expect the key to be available since we execute the put later than the delete. Yet, I'm fairly sure the key will not exist since on commit we will first process all the puts, and after that all the deletes.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/IdleStateMonitor.java
```diff
@@ -0,0 +1,74 @@
...
+
+import io.camunda.zeebe.engine.processing.streamprocessor.StreamProcessorListener;
+import io.camunda.zeebe.engine.processing.streamprocessor.TypedRecord;
+import io.camunda.zeebe.logstreams.log.LogStreamReader;
+import io.camunda.zeebe.logstreams.log.LoggedEvent;
+import io.camunda.zeebe.logstreams.storage.LogStorage;
+import java.util.ArrayList;
+import java.util.List;
+
+final class IdleStateMonitor implements LogStorage.CommitListener, StreamProcessorListener {
```
remcowesterhoud, line 11: ❌ We should write some unit tests for this logic

remcowesterhoud (commented):
@pihme The changes look good! 
I've added some comments. Please take a look at them once you have time.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/db/InMemoryZeebeDbColumnFamilyIterationContext.java
```diff
@@ -0,0 +1,43 @@
...
+ * This class allows iterating over a subset of keys in the database. The subset is identified by a
+ * common prefix for all keys in that subset. It also implements a recursion guard by checking that
+ * iterations can only be nested up to a certain maximum depth.
+ */
+final class InMemoryZeebeDbColumnFamilyIterationContext {
+
+  private final long columnFamilyPrefix;
+
+  private final Queue<ExpandableArrayBuffer> prefixKeyBuffers =
+      new ArrayDeque<>(List.of(new ExpandableArrayBuffer(), new ExpandableArrayBuffer()));
```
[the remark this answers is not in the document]
pihme: This is a strange this. I asked Chris about it, but I am not really any wiser. Basically it is a reentry/recursion protection. With two buffers you can have two nested iterations. Why it is two and not any other number is not clear to me. It is compatible with EZE though

src/test/java/io/camunda/zeebe/bpmnassert/testengine/db/InMemoryZeebeDbTransactionTest.java
pihme: Pretty much just copy and pasted it from EZE, tbh.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/db/InMemoryZeebeDbTransactionContext.java
pihme: Again, copied it from EZE. I agree with your point though. It would make more sense to call e.g. `transaction.reset();` (which internally does a rollback :-)

src/test/java/io/camunda/zeebe/bpmnassert/testengine/db/InMemoryZeebeDbTransactionTest.java
pihme: Nice one. If that fails, we should also tell Chris and Philipp, cause this may also happen in EZE.

pihme:
@remcowesterhoud Looked over all your comments. Feel free to change things. I will focus on medic duty this week and don't want to stall progress.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/db/InMemoryZeebeDbTransactionContext.java
remcowesterhoud: How would you feel about just changing the `finally` into a `catch`? `commit()` already clears the caches after applying them to the database so it wouldn't be necessary to have a finally.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/RecordStreamSource.java
remcowesterhoud: `RecordStreamSource` is an interface, so I can not add this here. I'd have to add it to `ZeebeEngineImpl` but it doesn't feel right to have the `toString` of the complete engine only print out the records.

pihme: You can add a `default toString()` 

And I think I had a different comment that in my mind `ZeebeEngine` should not implement `RecordStreamSource` but rather have a getter to return a record stream source. So those two go hand in hand, in my mind.

src/test/java/io/camunda/zeebe/bpmnassert/testengine/PrintRecordStreamExtension.java
remcowesterhoud: I copied this from EZE. I think it makes sense because this is only used for the unit tests of the engine. EZE has a similar construct in the extension and I think we should add this to our extension as well.

When we split this project into multiple modules our engine will need this anyway as it shouldn't depend on the module that contains our extension.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/IdleStateMonitor.java
remcowesterhoud: Unfortunately the `IdleStateMonitor` is not stable as it is now. I tried writing unit tests to verify the waiting for idle state works correctly, but they kept failing.
I propose we merge this PR first and have another look at this separate from this PR.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/RecordStreamSource.java
remcowesterhoud: It's still a bit awkward to do this. The purpose of `print()` is to write the log to the console. It doesn't return anything. `toString()` should be a string representation of the object. I think returning a string containing all log records is a bit excessive.

remcowesterhoud:
@pihme I've processed all of the review comments. I've split them up by commit so it's easier to review them. If you find the time please have another look 😁

src/test/java/io/camunda/zeebe/bpmnassert/testengine/EngineClientTest.java
```diff
@@ -98,7 +101,7 @@ public void shouldRequestTopology() {
   @Test
   public void shouldUseBuiltInClient() {
     // given
-    final ZeebeClient zeebeClient = zeebeEngine.createClient();
+    zeebeClient = zeebeEngine.createClient();
```
pihme: :x: I think it would be better not to overwrite the class field. Instead, we could use a different name here and close it within the test. And then `zeebeClient` is not touched at all, but opened and closed by the general code.

Same for other instances in this class, where the field gets overwritten

src/main/java/io/camunda/zeebe/bpmnassert/testengine/RecordStreamSourceImpl.java
```diff
@@ -24,26 +30,31 @@
...
-    this.records = records;
+  public RecordStreamSourceImpl(final LogStreamReader logStreamReader, final int partitionId) {
+    this.logStreamReader = logStreamReader;
+    this.partitionId = partitionId;
+  }
+
+  @Override
+  public Iterable<Record<?>> records() {
+    updateWithNewRecords();
+    return records;
```
pihme, line 46: :x: Please return `Collections.unmodifiableList(records)` or something like this. Now that the records get updated, we need to protect the list against modification from the outside.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/RecordStreamSourceImpl.java
```diff
@@ -122,4 +133,45 @@ public void print(final boolean compact) {
       System.out.println("---------------------------");
     }
   }
+
+  private void updateWithNewRecords() {
+    if (lastPosition < 0) {
+      logStreamReader.seekToFirstEvent();
+    } else {
+      logStreamReader.seekToNextEvent(lastPosition);
```
pihme, line 141: :question: Does this work as expected when there is no event after last position?

src/main/java/io/camunda/zeebe/bpmnassert/testengine/ZeebeEngineImpl.java
```diff
@@ -16,7 +17,7 @@
 
 public class ZeebeEngineImpl implements ZeebeEngine {
 
-  public static final int PORT = 26500;
+  public static final int PORT = SocketUtil.getNextAddress().getPort();
```
pihme: :x: this cannot be static if we want to have different ports for different instances

src/main/java/io/camunda/zeebe/bpmnassert/testengine/PartitionCommandSenderImpl.java
```diff
@@ -8,28 +8,18 @@
...
-    // delegate the command to the subscription handler of the receiver partition
-    try {
-      subscriptionHandlers.get(receiverPartitionId).apply(bytes);
-    } catch (NullPointerException npe) {
-      throw new RuntimeException(
-          String.format(
-              "No message subscription handler found for partition `%s`", receiverPartitionId));
-    }
-
+    subscriptionHandler.apply(bytes);
```
pihme, line 22: :wrench: maybe add a check to this method in the beginning that the partition is what we expect it to be, and throw an exception otherwise. Without that I feel it is not clear that we are doing a deliberate modification here.

Dito in other places where it applies

src/test/java/io/camunda/zeebe/bpmnassert/testengine/db/InMemoryZeebeDbTransactionTest.java
```diff
@@ -468,6 +468,27 @@ public void shouldNotCommitOnError() {
     assertThat(threeColumnFamily.exists(threeKey)).isFalse();
   }
 
+  @Test
+  public void shouldWriteKeyAfterDeletion() {
```
pihme, line 472: :wrench: If you have time please, also create a PR in EZE repository. I would like to give back some code improvements where possible, as a thank you for letting us copy their stuff.

pihme (changes requested):
Great work! Thanks a lot!

I have a couple of remaining changes. 

Happy to look at it again, if you want. But you can also just merge it.

src/main/java/io/camunda/zeebe/bpmnassert/testengine/RecordStreamSourceImpl.java
remcowesterhoud: It won't throw any errors or that sort of thing. It returns `false` if there are no events after this position. This means `logStreamReader.hasNext()` resolves to `false` as well and the list of records doesn't get updated as there are no new records.

PR #83, opened 2021-12-01
Switch to our own in memory engine

remcowesterhoud:
## Description

- Added the injection of the engine, client en recordstream to the `ZeebeAssertionsExtension`
- Removed the EZE dependency and added Zeebe + Log4J dependencies
- Fixed all the unit tests so that they can run with the new engine

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/main/java/io/camunda/zeebe/bpmnassert/extensions/ZeebeAssertionsExtension.java
+++ b/src/main/java/io/camunda/zeebe/bpmnassert/extensions/ZeebeAssertionsExtension.java
@@ -13,9 +13,11 @@ import java.util.stream.Collectors;
 import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.TestWatcher;
 import org.junit.platform.commons.util.ReflectionUtils;
 
-public class ZeebeAssertionsExtension implements BeforeEachCallback, AfterEachCallback {
+public class ZeebeAssertionsExtension implements BeforeEachCallback, AfterEachCallback,
+    TestWatcher {
 
   private static final String KEY_ZEEBE_CLIENT = "ZEEBE_CLIENT";
   private static final String KEY_ZEEBE_ENGINE = "ZEEBE_ENGINE";
@@ -45,6 +47,15 @@ public class ZeebeAssertionsExtension implements BeforeEachCallback, AfterEachCa
     final Object engineContent = getStore(extensionContext).get(KEY_ZEEBE_ENGINE);
     final InMemoryEngine engine = (InMemoryEngine) engineContent;
     engine.stop();
+  }
+
+  @Override
+  public void testFailed(final ExtensionContext extensionContext, final Throwable cause) {
+    final Object engineContent = getStore(extensionContext).get(KEY_ZEEBE_ENGINE);
+    final InMemoryEngine engine = (InMemoryEngine) engineContent;
+
+    System.out.println("===== Test failed! Printing records from the stream:");
+    engine.getRecordStream().print(true);
   }
 
   private void injectFields(final ExtensionContext extensionContext, final Object... objects) {
--- a/src/main/java/io/camunda/zeebe/bpmnassert/extensions/ZeebeAssertionsExtension.java
+++ b/src/main/java/io/camunda/zeebe/bpmnassert/extensions/ZeebeAssertionsExtension.java
@@ -16,8 +16,8 @@ import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.TestWatcher;
 import org.junit.platform.commons.util.ReflectionUtils;
 
-public class ZeebeAssertionsExtension implements BeforeEachCallback, AfterEachCallback,
-    TestWatcher {
+public class ZeebeAssertionsExtension
+    implements BeforeEachCallback, AfterEachCallback, TestWatcher {
 
   private static final String KEY_ZEEBE_CLIENT = "ZEEBE_CLIENT";
   private static final String KEY_ZEEBE_ENGINE = "ZEEBE_ENGINE";
--- a/src/main/java/io/camunda/zeebe/bpmnassert/extensions/ZeebeAssertionsExtension.java
+++ b/src/main/java/io/camunda/zeebe/bpmnassert/extensions/ZeebeAssertionsExtension.java
@@ -1,11 +1,15 @@
 package io.camunda.zeebe.bpmnassert.extensions;
 
 import io.camunda.zeebe.bpmnassert.RecordStreamSourceStore;
+import io.camunda.zeebe.bpmnassert.testengine.EngineFactory;
+import io.camunda.zeebe.bpmnassert.testengine.InMemoryEngine;
+import io.camunda.zeebe.bpmnassert.testengine.RecordStreamSource;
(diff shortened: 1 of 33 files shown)


src/main/java/io/camunda/zeebe/bpmnassert/extensions/ZeebeAssertionsExtension.java
```diff
@@ -1,77 +1,98 @@
...
-import org.camunda.community.eze.RecordStreamSource;
+import java.util.stream.Collectors;
 import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.TestWatcher;
 import org.junit.platform.commons.util.ReflectionUtils;
 
-public class ZeebeAssertionsExtension implements BeforeEachCallback, AfterEachCallback {
+public class ZeebeAssertionsExtension
```
pihme, line 19: :wrench: Maybe reconsider the name.

src/main/java/io/camunda/zeebe/bpmnassert/extensions/ZeebeAssertionsExtension.java
```diff
@@ -1,77 +1,98 @@
...
-    final ZeebeClient zeebeClient =
-        (ZeebeClient) zeebeClientField.get(extensionContext.getRequiredTestInstance());
-    getStore(extensionContext).put(KEY_ZEEBE_CLIENT, zeebeClient);
+  private Optional<Field> getField(final Field[] declaredFields, final Object object) {
+    final List<Field> fields =
+        Arrays.stream(declaredFields)
+            .filter(field -> field.getType().isInstance(object))
+            .collect(Collectors.toList());
+
+    if (fields.size() > 1) {
```
pihme, line 76: :question: Why is this important? We could inject it in all fields anyway

src/main/java/io/camunda/zeebe/bpmnassert/extensions/ZeebeAssertionsExtension.java
```diff
@@ -1,77 +1,98 @@
...
-                "Expected a field of type ZeebeClient to be declared in the test class, "
-                    + "but none has been found. Please make sure a field of type ZeebeClient "
-                    + "has been declared in the test class."));
+  private void injectField(
+      final ExtensionContext extensionContext, final Field field, final Object object) {
+    try {
+      ReflectionUtils.makeAccessible(field);
+      field.set(extensionContext.getRequiredTestInstance(), object);
+    } catch (IllegalAccessException e) {
+      // TODO proper error handling
```
pihme, line 93: :clown_face:

src/test/java/io/camunda/zeebe/bpmnassert/assertions/JobAssertTest.java
```diff
@@ -28,13 +31,12 @@
   public static final String ERROR_CODE = "error";
   public static final String ERROR_MSG = "error occurred";
 
-  private ZeebeEngine engine;
-
   // These tests are for testing assertions as well as examples for users
   @Nested
   class HappyPathTests {
     private ZeebeClient client;
     private RecordStreamSource recordStreamSource;
```
pihme, line 38: Do we still need this field now?

src/test/java/io/camunda/zeebe/bpmnassert/util/Utilities.java
```diff
@@ -146,22 +149,26 @@ public static PublishMessageResponse sendMessage(
...
-    if (lastRecord != null) {
+      final InMemoryEngine engine, final ZeebeClient client, final String elementId) {
+    final List<Record<JobRecordValue>> records =
+        StreamFilter.jobRecords(engine.getRecordStream())
+            .withElementId(elementId)
+            .withIntent(JobIntent.CREATED)
+            .stream()
+            .collect(Collectors.toList());
+
+    if (!records.isEmpty()) {
```
pihme, line 166: :x: I think we need some kind of error handling or feedback so that callers of this method know whether it succeeded or not.

src/test/java/io/camunda/zeebe/bpmnassert/util/Utilities.java
```diff
@@ -146,22 +149,26 @@ public static PublishMessageResponse sendMessage(
...
   public static void completeTask(
-      final ZeebeEngine engine, final ZeebeClient client, final String elementId) {
-    Record<JobRecordValue> lastRecord = null;
-    for (final Record<JobRecordValue> record : engine.jobRecords().withElementId(elementId)) {
-      if (record.getIntent().equals(JobIntent.CREATED)) {
-        lastRecord = record;
-      }
-    }
-    if (lastRecord != null) {
+      final InMemoryEngine engine, final ZeebeClient client, final String elementId) {
```
pihme, line 158:
```suggestion
      final InMemoryEngine engine, final ZeebeClient client, final String serviceTaskElementId) {
```

src/test/java/io/camunda/zeebe/bpmnassert/util/Utilities.java
```diff
@@ -146,22 +149,26 @@ public static PublishMessageResponse sendMessage(
...
+    final List<Record<JobRecordValue>> records =
+        StreamFilter.jobRecords(engine.getRecordStream())
+            .withElementId(elementId)
+            .withIntent(JobIntent.CREATED)
+            .stream()
+            .collect(Collectors.toList());
+
+    if (!records.isEmpty()) {
+      final Record<JobRecordValue> lastRecord;
+      lastRecord = records.get(records.size() - 1);
```
pihme, line 168: :x: Not clear to me why it would pick the last record. I would expect either the first (FIFO) or it completes all of them

pihme (changes requested):
Looking forward to this. A few smaller minor changes would round it off.

src/main/java/io/camunda/zeebe/bpmnassert/extensions/ZeebeAssertionsExtension.java
remcowesterhoud: We could. I just thought it might be confusing from a user perspective if he defines two clients that they will both be the same client. And I don't think we should manage multiple clients in our extension.

I don't see any reason why anyone would need the fields twice so I decided prevent people from doing this and causing unexpected behaviours.

src/test/java/io/camunda/zeebe/bpmnassert/assertions/JobAssertTest.java
remcowesterhoud: No, I am gonna do a refactor of the tests and get rid of fields that aren't used anymore. I just didn't want this PR to become too big.

src/test/java/io/camunda/zeebe/bpmnassert/util/Utilities.java
remcowesterhoud: I just copied what we already had tbh. I think it is the last in case there is a loop. Than you wouldn't want to get the job you already completed earlier.

The real problem is that it uses this "hacky" and completes jobs without activating them first. I'll keep it like this for now, we can do it in a proper way when we implement the `TestClient`

remcowesterhoud:
Thanks for the review @pihme. I've responded / fixed your suggestions.

I've renamed `@ZeebeAssertions` to `@ZeebeProcessTest`. Let me know what you think of this name!

pihme (approved):
Love it!

PR #101, opened 2021-12-08
Rename project to zeebe-process-test

remcowesterhoud:
## Description

I recommend reviewing by commit

We have decided on a definitive name for the project. We went with `zeebe-process-test`. This is because it does not contain BPMN, for when we want to support DMN in the future. It also doesn't contain assertion which caused confusion, as the project provided more than just assertions.

- Renamed the artifact (also in README)
- Renamed the scm settings
- Renamed the modules

This repository has also been renamed. This means your git remote url should also be updated!

We still need to rename some stuff in the GitHub workflows. This will be done in a later PR as this needs te done together with the infra team.

## Definition of Done

Code changes:

Testing:

Documentation:


README.md
```diff
@@ -16,7 +16,7 @@ Add the following dependency to your project
 ```xml
 <dependency>
   <groupId>io.camunda</groupId>
-  <artifactId>zeebe-bpmn-assert</artifactId>
+  <artifactId>zeebe-process-test</artifactId>
```
pihme, line 19: :x: Please also consider renaming top level heading in Readme. And maybe also rephrasing the first paragraph Camunda Cloud vs. Zeebe. 

:x: There is also this line in the readme.md which needs to be updated
> For example tests the best place to look right now is the tests in io.camunda.zeebe.bpmnassert.assertions

And maybe do a general text search for the old name to see if anything else remained

pihme (changes requested):
Looks good. Some details are missing.

README.md
pihme: :clown_face: Love the avatar, by the way :clown_face:

remcowesterhoud: Thanks 😄 

I've fixed the README. The general text search yielded no results apart from the Vault which I mentioned before. That'll probably be fixed in a separate PR tomorrow.

pihme (approved):

PR #114, opened 2021-12-20
Make RecordStreamSourceImpl threadsafe

### LINKED ISSUE #110: Occasionally weird errors like 'Unable to determine array type, found unknown header byte 0xc4 at reader offset 59'
## Description
I am currently playing with spring-zeebe to support testing in spring environments based on zeebe-process-test 1.3.0-alpha3. I occasionally get the below exception on an assertion of process completion `assertThat(processInstance).isCompleted()`. Could it have to do with multi-threading, that some thread is still executing jobs and I do an assertion at the same time?

```
java.lang.RuntimeException: Could not deserialize object [JobBatchRecord]. Deserialization stuck at offset 60 of length 201

	at io.camunda.zeebe.msgpack.UnpackedObject.wrap(UnpackedObject.java:38)
	at io.camunda.zeebe.logstreams.impl.log.LoggedEventImpl.readValue(LoggedEventImpl.java:135)
	at io.camunda.zeebe.process.test.testengine.RecordStreamSourceImpl.mapToRecord(RecordStreamSourceImpl.java:169)
	at io.camunda.zeebe.process.test.testengine.RecordStreamSourceImpl.updateWithNewRecords(RecordStreamSourceImpl.java:148)
	at io.camunda.zeebe.process.test.testengine.RecordStreamSourceImpl.records(RecordStreamSourceImpl.java:46)
	at io.camunda.zeebe.process.test.testengine.RecordStreamSourceImpl.recordsOfValueType(RecordStreamSourceImpl.java:53)
	at io.camunda.zeebe.process.test.testengine.RecordStreamSourceImpl.processInstanceRecords(RecordStreamSourceImpl.java:60)
	at io.camunda.zeebe.process.test.filters.StreamFilter.processInstance(StreamFilter.java:9)
	at io.camunda.zeebe.process.test.assertions.ProcessInstanceAssert.isProcessInstanceCompleted(ProcessInstanceAssert.java:106)
	at io.camunda.zeebe.process.test.assertions.ProcessInstanceAssert.isCompleted(ProcessInstanceAssert.java:82)
	at io.berndruecker.playground.zeebe.tests.SimpleSpringBootTest.lambda$waitForCompletion$0(SimpleSpringBootTest.java:77)
	at org.testcontainers.shaded.org.awaitility.core.AssertionCondition.lambda$new$0(AssertionCondition.java:53)
	at org.testcontainers.shaded.org.awaitility.core.ConditionAwaiter$ConditionPoller.call(ConditionAwaiter.java:234)
	at org.testcontainers.shaded.org.awaitility.core.ConditionAwaiter$ConditionPoller.call(ConditionAwaiter.java:221)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:831)
Caused by: java.lang.RuntimeException: Could not read property 'variables'
	at io.camunda.zeebe.msgpack.value.ObjectValue.read(ObjectValue.java:112)
	at io.camunda.zeebe.msgpack.UnpackedObject.wrap(UnpackedObject.java:32)
	... 17 more
Caused by: io.camunda.zeebe.msgpack.spec.MsgpackReaderException: Unable to determine array type, found unknown header byte 0xc4 at reader offset 59
	at io.camunda.zeebe.msgpack.spec.MsgPackReader.exceptionOnUnknownHeader(MsgPackReader.java:474)
	at io.camunda.zeebe.msgpack.spec.MsgPackReader.readArrayHeader(MsgPackReader.java:107)
	at io.camunda.zeebe.msgpack.value.ArrayValue.read(ArrayValue.java:94)
	at io.camunda.zeebe.msgpack.property.BaseProperty.read(BaseProperty.java:75)
	at io.camunda.zeebe.msgpack.value.ObjectValue.read(ObjectValue.java:110)
	... 18 more
```

@remcowesterhoud maybe?

remcowesterhoud:
## Description

I've added a synchronized around the updating of the list of records in the record stream. I've written a test to verify this change solved the issue in #110 

## Related issues

closes #110 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/main/java/io/camunda/zeebe/process/test/testengine/RecordStreamSourceImpl.java
+++ b/src/main/java/io/camunda/zeebe/process/test/testengine/RecordStreamSourceImpl.java
@@ -33,8 +33,8 @@ public class RecordStreamSourceImpl implements RecordStreamSource {
 
   private final LogStreamReader logStreamReader;
   private final int partitionId;
-  private List<Record<?>> records = new ArrayList<>();
-  private long lastPosition = -1L;
+  private final List<Record<?>> records = new ArrayList<>();
+  private volatile long lastPosition = -1L;
 
   public RecordStreamSourceImpl(final LogStreamReader logStreamReader, final int partitionId) {
     this.logStreamReader = logStreamReader;
@@ -137,17 +137,19 @@ public class RecordStreamSourceImpl implements RecordStreamSource {
   }
 
   private void updateWithNewRecords() {
-    if (lastPosition < 0) {
-      logStreamReader.seekToFirstEvent();
-    } else {
-      logStreamReader.seekToNextEvent(lastPosition);
-    }
+    synchronized (logStreamReader) {
+      if (lastPosition < 0) {
+        logStreamReader.seekToFirstEvent();
+      } else {
+        logStreamReader.seekToNextEvent(lastPosition);
+      }
 
-    while (logStreamReader.hasNext()) {
-      final LoggedEvent event = logStreamReader.next();
-      final CopiedRecord<UnifiedRecordValue> record = mapToRecord(event);
-      records.add(record);
-      lastPosition = event.getPosition();
+      while (logStreamReader.hasNext()) {
+        final LoggedEvent event = logStreamReader.next();
+        final CopiedRecord<UnifiedRecordValue> record = mapToRecord(event);
+        records.add(record);
+        lastPosition = event.getPosition();
+      }
     }
   }
 
--- a/src/test/java/io/camunda/zeebe/process/test/multithread/MultiThreadTest.java
+++ b/src/test/java/io/camunda/zeebe/process/test/multithread/MultiThreadTest.java
@@ -1,0 +1,69 @@
+package io.camunda.zeebe.process.test.multithread;
+
+import static io.camunda.zeebe.process.test.assertions.BpmnAssert.assertThat;
+import static io.camunda.zeebe.process.test.util.Utilities.deployProcess;
+import static io.camunda.zeebe.process.test.util.Utilities.startProcessInstance;
+
+import io.camunda.zeebe.client.ZeebeClient;
+import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
+import io.camunda.zeebe.process.test.RecordStreamSourceStore;
+import io.camunda.zeebe.process.test.extensions.ZeebeProcessTest;
+import io.camunda.zeebe.process.test.testengine.InMemoryEngine;
+import io.camunda.zeebe.process.test.testengine.RecordStreamSource;
+import io.camunda.zeebe.process.test.util.Utilities.ProcessPackStartEndEvent;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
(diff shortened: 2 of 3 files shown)


src/main/java/io/camunda/zeebe/process/test/testengine/RecordStreamSourceImpl.java
```diff
@@ -33,8 +33,8 @@
 
   private final LogStreamReader logStreamReader;
   private final int partitionId;
-  private List<Record<?>> records = new ArrayList<>();
-  private long lastPosition = -1L;
+  private volatile List<Record<?>> records = new ArrayList<>();
```
pihme, line 36: :x:
```suggestion
  private final List<Record<?>> records = new ArrayList<>();
```
This field is never written to.

src/test/java/io/camunda/zeebe/process/test/multithread/MultiThreadTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+@ZeebeProcessTest
+public class MultiThreadTest {
```
pihme, line 25: :thought_balloon: I am of two minds about this test. On the one hand I am glad there is a test at all. But this test looks like it might be a flaky test. It feels too short, and I could imagine even with the wrong implementation it will sometimes pass and sometimes fail. I might be wrong though. 

Also, I don't really have a better idea.

There is the option of using _jqwik_ and having a number of runs with a random number of runners and each looping a random number of times with a random operation. But at the same time this feels like overkill. 

Still, if you feel like making this more solid, you could have a look at: https://github.com/camunda-cloud/zeebe/blob/develop/atomix/cluster/src/test/java/io/atomix/raft/RandomizedRaftTest.java or https://github.com/camunda-cloud/zeebe/blob/develop/broker/src/test/java/io/camunda/zeebe/broker/system/partitions/impl/RandomizedPartitionTransitionTest.java

None of them are clean examples of property based tests. I am not saying you need to change this test. But if you either feel uncomfortable about this test and/or are curious about property based testing, you might give this a go

pihme (approved):
Looks good! :tada: 

There is one minor change. If you only change that, I don't need to see it again before merge.
There is also a suggestion for improving the test. If you follow that suggestion, I would like to see the changed test again. 

Also happy with merging it now, releasing the bugfix and having a follow up issue to improve the test.

src/test/java/io/camunda/zeebe/process/test/multithread/MultiThreadTest.java
remcowesterhoud: I understand your concern. However, the test didn't seem flaky when I was playing around with it. Without the fix it failed 100% of the runs, and with the fix I haven't seen it fail yet.

I would propose to keep it as is and if it turns out to be flaky and becomes a problem we can fix it at that time.

PR #137, opened 2022-01-12
Publish test report upon failure in deploy workflow

remcowesterhoud:
## Description

The deploy-artifact workflow did not publish the unit test results. In case of a test failure you'd have to manually go through the logs to see what failed. Adding these steps will publish the results to make it easier to see what went wrong.

Tested manually: https://github.com/camunda-cloud/zeebe-process-test/actions/runs/1687648571

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/.github/workflows/deploy-artifact.yml
+++ b/.github/workflows/deploy-artifact.yml
@@ -80,3 +80,18 @@ jobs:
           asset_path: ${{ steps.release.outputs.artifacts_archive_path }}
           asset_name: ${{ steps.release.outputs.artifacts_archive_path }}
           asset_content_type: application/zip
+
+      - name: Archive Test Results on Failure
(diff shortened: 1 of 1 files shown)


pihme (approved):
Looks good to me.

The only thing is that it has now more overlap with `build-test.yml`. Naively, I would think that the two workflows do different things. Which they do, in a way. But another difference right now seems to be that they also have different triggers and then do similar things.

Nothing to do right now, but maybe the two worklfows could be merged at some point

remcowesterhoud:
I'm hoping it'll be part of the community release action soon (https://github.com/camunda-community-hub/community-action-maven-release/issues/34). After that it could still be combined, but it would be way less duplicated.

PR #143, opened 2022-01-13
Build and run tests on different OS's

### LINKED ISSUE #142: Run test on multiple OS's in workflow
**Description**

In #141 we fixed an issue where tests failed on different OS's because of line separators. By modifying the workflow to run the tets on different OS's we can catch bugs like this before they cause problems.

remcowesterhoud:
## Description

Run the unit tests on different Operating Systems so we find OS specific bugs early.

## Related issues

closes #142

## Definition of Done

Code changes:

Testing:

Documentation:


src/test/java/io/camunda/zeebe/process/test/util/Utilities.java
```diff
@@ -152,7 +152,7 @@ public static void increaseTime(final InMemoryEngine engine, final Duration dura
     engine.increaseTime(duration);
     try {
       // we need to wait some physical time so that InMemoryEngine has a chance to fire the timers
-      Thread.sleep(50);
+      Thread.sleep(100);
```
remcowesterhoud, line 155: I know you will probably comment on this, so let me say in advance that this is temporary until I have a better solution. It's also not part of a public interface or anything, purely used for our own unit tests. 😄

korthout: LOL 😆 

On a serious note, thanks for pointing this out. That's very useful 👍

.github/workflows/build-test.yml
```diff
@@ -25,19 +31,30 @@ jobs:
...
 
       - name: Archive Test Results on Failure
         uses: actions/upload-artifact@v2
-        if: failure()
+        if: always()
         with:
-          name: test-results
+          name: Unit Test Results ${{ matrix.os }}
           path: target/surefire-reports/
           retention-days: 7
```
korthout, line 42: ❓ Previously this artifact was only uploaded on failures, why does it have to happen as well for success? Is it so that the test results can be published or so? It's probably unclear to me because they use different paths.

.github/workflows/build-test.yml
```diff
@@ -25,19 +31,30 @@ jobs:
...
 
-      - name: Publish Unit Test Results
-        id: publish
-        uses: EnricoMi/publish-unit-test-result-action@v1.26
-        if: failure()
-        with:
-          files: target/surefire-reports/*.xml
+  publish-test-results:
+      name: Publish Unit Test Results
+      needs: build-and-test
```
korthout, line 46: 🔧  I was a bit confused about the needs here, because that requires the referred job to have succeeded. It would be nice to have it directly next to the `if: always()`

korthout (approved):
LGTM 👍  I do have a question, but that's not blocking

Issue #158, opened 2022-01-31
Exceptions when using current test library within spring zeebe


berndruecker:
I create an in-memory engine in Spring Zeebe with the latest 1.3.2 version, when registering a worker I get this exception.

However, it also seems to be a timing issue of some sort, with re-arranging some spring code I also got around that exception. But I still did not figure out the exact why...

Here is the log - the sequence looks good to me:
```
2022-01-31 10:02:26.706  INFO 31280 --- [           main] s.c.c.ZeebeTestClientSpringConfiguration : Create Zeebe in-memory engine for test run...
2022-01-31 10:02:26.709  WARN 31280 --- [           main] io.camunda.zeebe.test.util               : No system property 'testForkNumber' set, using default value 0
2022-01-31 10:02:26.710  WARN 31280 --- [           main] io.camunda.zeebe.test.util               : No system property 'testMavenId' set, using default value 0
2022-01-31 10:02:26.710  INFO 31280 --- [           main] io.camunda.zeebe.test.util.SocketUtil    : Starting socket assignment with testForkNumber 0 and testMavenId 0
2022-01-31 10:02:26.712  INFO 31280 --- [           main] io.camunda.zeebe.test.util.SocketUtil    : Choosing next port 1025 for test fork 0 with range PortRange{host='localhost', basePort=1025, maxOffset=100, currentOffset=1}
2022-01-31 10:02:27.396  INFO 31280 --- [   -zb-actors-3] org.camunda.feel.FeelEngine              : Engine created. [value-mapper: CompositeValueMapper(List(io.camunda.zeebe.el.impl.feel.MessagePackValueMapper@42114dfe)), function-provider: io.camunda.zeebe.el.impl.feel.FeelFunctionProvider@23dfaf20, clock: io.camunda.zeebe.el.impl.ZeebeFeelEngineClock@343f683a, configuration: Configuration(false)]
2022-01-31 10:02:27.500  INFO 31280 --- [   -zb-actors-3] io.camunda.zeebe.logstreams              : Recovered state of partition 1 from snapshot at position -1
[...]

PR #160, opened 2022-02-01
Create separate stream writers and don't write synchronously

remcowesterhoud:
## Description

Running tests would sometimes result in an IndexOutOfBoundsException upon writing commands. This is most likely caused by multiple threads using the same writer. The first thread would be writing, while the seconds thread resets the writer.

## Related issues

Related to #158 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/src/main/java/io/camunda/zeebe/process/test/testengine/GrpcToLogStreamGateway.java
+++ b/src/main/java/io/camunda/zeebe/process/test/testengine/GrpcToLogStreamGateway.java
@@ -94,16 +94,20 @@ public class GrpcToLogStreamGateway extends GatewayGrpc.GatewayImplBase implemen
 
   private void writeCommandWithKey(
       final Long key, final RecordMetadata metadata, final BufferWriter bufferWriter) {
-    writer.reset();
+    synchronized (writer) {
+      writer.reset();
 
-    writer.key(key).metadataWriter(metadata).valueWriter(bufferWriter).tryWrite();
+      writer.key(key).metadataWriter(metadata).valueWriter(bufferWriter).tryWrite();
+    }
   }
 
   private void writeCommandWithoutKey(
       final RecordMetadata metadata, final BufferWriter bufferWriter) {
-    writer.reset();
+    synchronized (writer) {
+      writer.reset();
 
-    writer.keyNull().metadataWriter(metadata).valueWriter(bufferWriter).tryWrite();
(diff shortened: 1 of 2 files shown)


remcowesterhoud:
@Zelldon Would you mind reviewing this? I feel like you know most about the `InMemoryEngine` to know if this change makes sense

src/main/java/io/camunda/zeebe/process/test/testengine/GrpcToLogStreamGateway.java
```diff
@@ -94,16 +94,20 @@ public GrpcToLogStreamGateway(
 
   private void writeCommandWithKey(
       final Long key, final RecordMetadata metadata, final BufferWriter bufferWriter) {
-    writer.reset();
+    synchronized (writer) {
```
Zelldon, line 97: :x: This is not necessary. It is accessed by one single thread.

Zelldon (approved):
Makes sense to me, to use separate writers for different actors/threads. The synchronize is in my opinion not necessary.

I had loved we would had one base (eze) and we could fix it in one place. :shrug:

src/main/java/io/camunda/zeebe/process/test/testengine/GrpcToLogStreamGateway.java
remcowesterhoud: Is it? Because after I added this I have not seen the issue anymore. So I was under the impression that this was the most important part of the fix.

Zelldon: Maybe I oversee something but these are private methods which are always called in the executor, which is single threaded.

remcowesterhoud: You're right, I can't figure out why but I tested it again and this clearly fixes the issue. I'm gonna merge it and dig deeper into the why this would change anything later.

Zelldon: @remcowesterhoud I think the problem is this method https://github.com/camunda-cloud/zeebe-process-test/blame/a10f59519ab4440f2f76d0e2156086c87fa9e7c3/src/main/java/io/camunda/zeebe/process/test/testengine/GrpcToLogStreamGateway.java#L284-L302

PR #166, opened 2022-02-03
Split engine and assertions into separate modules

### LINKED ISSUE #90: Split in separate modules
**Description**

As it is now everything is part of a single module. We should split this into separate modules. This way we can separate the engine from the rest of the code, which prepares us to run the engine in a testcontainer and allows users to use different java versions from the engine.

remcowesterhoud:
## Description

The first step to supporting Java 8 is isolating the engine from the assertions. With this change I have split the code into 2 separate modules, the engine module and the assertions module.

The engine module is responsible for everything that has to do with running the processes. It will house the `InMemoryEngine` and the `InMemoryDb`. This module will be dependent on other Zeebe modules and will stay up to date with Java versions.

The assertions module is responsible for verifying the process. It contains assertions and other methods to traverse the record stream in order to verify the process behaves as expected. It also contains the `ZeebeProcessTest` annotation.
This module should support Java 8. This means it cannot have any dependencies on Zeebe modules that will update to higher Java versions. For now it still has a dependency upon the engine module, but this will be removed once we can run the engine in a testcontainer.

## Related issues

closes #90 

## Definition of Done

Code changes:

Testing:

Documentation:


assertions/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.camunda</groupId>
+    <artifactId>zeebe-process-test-root</artifactId>
+    <version>1.3.4-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>zeebe-process-test</artifactId>
```
pihme, line 12: :thinking: not sure about the artifact name. Maybe invent a new name like `zeebe-process-test-assertions`.

pihme (commented):
So... this one is a little difficult to review. You say it's a first step. Maybe it's just different from the first step I would have made. Or maybe the first step is too small for me.

Let me tell you what I mean, and then maybe you can tell me what your next steps are going to be.

I guess my first step would have been to extract (and rename) `InMemoryEngine` and `RecordStreamSource` into their own module. This module will only have these two interfaces, most likely. And it's artifact name is probably something ending in `-api`.

Then both `*-test-engine` and `*-process-test` could depend on these shared interfaces to a) control the engine and b) fetch the records from it.

I am also still dwelling on the idea that I would at some point in the future like to see a module which has just the assertions, without the extension. I guess it is not strictly necessary since we decided we don't wont multi-version extension, but instead always use the containerized form. 

However, having the assertions cleanly without the extension would allow us to use them inside Zeebe to rewrite some of the tests which are based on records. Then we would just need to rewrite the logic to provide the records.

But if the extension stay part of it, then it will pull in additional dependencies to test containers and such things. But maybe I am getting ahead of myself.

Lastly, Nicolas was lately talking about using Flatten POM in Zeebe. I must admit I haven't fully understood how it works. But from what I gathered I think it could be a game changer to deal with transient dependencies. So I would encourage you to talk to Nicolas about it and employ it here, whenever you can.

remcowesterhoud:
Thanks for reviewing. I agree that it makes sense to move the interfaces to a separate module as well. The whole communication between the assertions and the testcontainer is still a bit unclear to me. Maybe we can have a chat about that soon?

I hear you about the extension. I wanted to keep this change "small" and figured we could do this in the future. But now that I think about it again, moving the extension would be a breaking change so it'd be better to do this right away.

I saw the post about flattening pom files. I want to look into it properly first, but from what I understood it's just a plugin. I'll create a new issue for it. I isn't part of the module split so I don't want to add it to this PR.

pihme:
Just had a look at it. I was wondering whether all tests need to be in QA. I would hope that some tests could just as well be in the module of the classes they are testing ( e.g. `EngineClientTest`, `...inspections.*` )

remcowesterhoud:
@pihme I don't agree on moving the `...inspections.*` tests. They are the same type of tests as the assertion tests in my opinion. The engine tests I have moved to the engine module.

I have also moved the `RecordStreamSourceStore` to make it part of `BpmnAssert`. I'd also like to remove the `InspectionUtility` and make the entry points of the inspections also part of `BpmnAssert` (not part of this PR). This would mean renaming `BpmnAssert` to something that's a bit more descriptive of what it does.

pihme (changes requested):
Looks good for the most part. Three comments for the last final touches.

Please also document the different modules and their Java version in `README.md`

pihme (changes requested):
Did you forget a commit? I cannot find the changes to the POM files which you marked as resolved.

remcowesterhoud:
I committed it to the wrong branch 🙈 It's pushed now.

PR #168, opened 2022-02-08
Add assertion entry point for ProcessInstanceResult

### LINKED ISSUE #167: Add ProcessInstanceResult entry point for ProcessInstanceAssertions
**Description**

Starting a process instance with result will return a `ProcessInstanceResult`. It should be possible to use this object as an entry point for the Process Instance assertions.

remcowesterhoud:
## Description

We already had an entry point for ProcessInstanceEvents, but we missed the ProcessInstanceResults. This change adds this entry point for allowing easy assertions when users have a ProcessInstanceResult.

## Related issues

closes #167 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/assertions/src/test/java/io/camunda/zeebe/process/test/assertions/BpmnAssertTest.java
+++ b/assertions/src/test/java/io/camunda/zeebe/process/test/assertions/BpmnAssertTest.java
@@ -1,0 +1,107 @@
+package io.camunda.zeebe.process.test.assertions;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.mockito.Mockito.mock;
+
+import io.camunda.zeebe.client.api.response.ActivatedJob;
+import io.camunda.zeebe.client.api.response.DeploymentEvent;
+import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
+import io.camunda.zeebe.client.api.response.ProcessInstanceResult;
+import io.camunda.zeebe.client.api.response.PublishMessageResponse;
+import io.camunda.zeebe.process.test.api.RecordStreamSource;
+import io.camunda.zeebe.process.test.inspections.model.InspectedProcessInstance;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+class BpmnAssertTest {
+
+  @BeforeEach
+  void beforeEach() {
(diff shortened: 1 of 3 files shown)


remcowesterhoud:
@saig0 Would you mind reviewing this one for me? It's a small change. Let me know if you have any questions!

assertions/src/test/java/io/camunda/zeebe/process/test/assertions/BpmnAssertTest.java
```diff
@@ -0,0 +1,100 @@
...
+    BpmnAssert.initRecordStream(mock(RecordStreamSource.class));
+  }
+
+  @AfterEach
+  public void afterEach() {
+    BpmnAssert.resetRecordStream();
+  }
+
+  @Test
+  public void testAssertThatProcessInstanceEventReturnsProcessInstanceAssert() {
```
saig0, line 30: :wrench: We could follow the convention of JUnit 5 and remove the `public` modifier of the test method.

Additionally, we could think about using `@DisplayName("Should return ProcessInstanceAssert for ProcessInstanceEvent")` to have human-friendly test names.

assertions/src/test/java/io/camunda/zeebe/process/test/assertions/BpmnAssertTest.java
```diff
@@ -0,0 +1,100 @@
...
+    // given
+    ProcessInstanceEvent event = mock(ProcessInstanceEvent.class);
+
+    // when
+    ProcessInstanceAssert assertions = BpmnAssert.assertThat(event);
+
+    //then
+    assertThat(assertions).isInstanceOf(ProcessInstanceAssert.class);
```
saig0, line 38: The test would not compile if `BpmnAssert.assertThat()` would return something else than `ProcessInstanceAssert`.

:wrench: We could modify the test to fail in the assertion instead of not compiling.

```suggestion
    final var assertions = BpmnAssert.assertThat(event);

    //then
    assertThat(assertions).isInstanceOf(ProcessInstanceAssert.class);
```

saig0 (approved):
@remcowesterhoud looks good :+1: 

I have a few comments that are completely optional. So, feel free to ignore.

assertions/src/test/java/io/camunda/zeebe/process/test/assertions/BpmnAssertTest.java
remcowesterhoud: Unfortunately we can't do this as this module should support Java 8

PR #177, opened 2022-02-10
Containerize in memory engine

### LINKED ISSUE #5: Make the engine runnable in a testcontainer
**Description**

As of now the engine needs to be embedded on the client side. This has its advantages, because it is very fast to start and stop the engine. However, using the engine as an embedded engine has the consequence that users can't use any Java version that's older that the Java version of the engine. Since the expectation is that some users will still be using Java 8 this would mean the engine can't upgrade to a version higher than 8. This is not what we want.

If we run the engine in a testcontainer we decouple this dependency, and we can upgrade the engine to whatever Java version we'd like.

We also need to open the debug port in the testcontainer and document how to connect to this debug port.

**Related issues**
Depends on #6
Depends on #24

remcowesterhoud:
## Description

1. Made the engine executable by adding a main class which creates and starts the engine. This required giving the gRPC server a static port, otherwise we won't know how to bind the containers ports later on. I've also created a Dockerfile to build the container.
2. Added a build step to the deploy-artifacts workflow which will build and push the container to docker hub.
3. Made the build-test workflow reusable, this way we can just call this workflow from the deploy-artifacts workflow and run all tests on different OSs, instead of relying on the community maven release plugin for testing.

## Related issues

closes #5 
Next steps will be done in #7 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/.github/workflows/deploy-artifact.yml
+++ b/.github/workflows/deploy-artifact.yml
@@ -4,7 +4,7 @@
 #        deploys a RELEASE with the selected version
 #        updates the project version by incrementing the patch version
 #        commits the version update change to the repository's default branch (main).
-name: Deploy artifacts with Maven
+name: Deploy artifacts
 on:
   push:
     branches: [main]
@@ -15,7 +15,7 @@ jobs:
   test:
     uses: ./.github/workflows/build-test.yml
 
-  build:
+  deploy-to-maven:
     runs-on: ubuntu-latest
     needs: [test]
     steps:
@@ -59,7 +59,7 @@ jobs:
           gpg_private_key: ${{ steps.secrets.outputs.MAVEN_CENTRAL_GPG_SIGNING_KEY_SEC }}
           passphrase: ${{ steps.secrets.outputs.MAVEN_CENTRAL_GPG_SIGNING_KEY_PASSPHRASE }}
 
-      - name: Deploy SNAPSHOT / Release
+      - name: Build and deploy to Maven
         id: release
         uses: camunda-community-hub/community-action-maven-release@v1.0.12
         with:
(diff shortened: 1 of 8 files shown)


.github/workflows/deploy-artifact.yml
```diff
@@ -83,3 +83,65 @@ jobs:
...
+          java-version: 11
+
+      - name: Build jar
+        run: |
+          cd engine
+          mvn clean package -DskipTests
+
+      - name: Build Docker image
+        run: |
+          if ! echo $TAG | grep SNAPSHOT; then
```
korthout, line 133: Interesting way to check the value

❓ is this a common way to check this?

If not, you might want to change it to `if [ "$TAG" != "SNAPSHOT" ]; then`, which I believe is the common way to check inequality of 2 strings in a shell script.

.github/workflows/deploy-artifact.yml
```diff
@@ -83,3 +83,65 @@ jobs:
...
+            echo ${{ github.event.release.tag_name }} >> $TAG
```
korthout, line 134: ❌ I don't think this does what you want. I think this doesn't replace the value of `TAG`. Please correct me if I'm wrong, perhaps the env-vars make this work differently than I expect. I would've expected something like `TAG="${{ github.event.release.tag_name }}"` for this to work.

korthout (changes requested):
Thanks @remcowesterhoud 🐳 

Generally looks good, but I have some questions.

.github/workflows/deploy-artifact.yml
remcowesterhoud: Your solution does something different than my solution does. Mine checks if `$TAG` contains the text `SNAPSHOT`. Yours checks if `$TAG` is not equal to `SNAPSHOT`. Generally `$TAG` will be filled with something like `1.0-SNAPSHOT` which means we don't want to enter the if-statement.

pom.xml
```diff
@@ -116,6 +116,12 @@
         <scope>import</scope>
       </dependency>
 
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-simple</artifactId>
+        <version>${dependency.slf4j.version}</version>
+      </dependency>
```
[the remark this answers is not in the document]
remcowesterhoud: Honestly, I think one of the other dependencies needs it. If I run the engine without this I won't get any logging, just see errors:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
SLF4J: Failed to load class "org.slf4j.impl.StaticMDCBinder".
SLF4J: Defaulting to no-operation MDCAdapter implementation.
SLF4J: See http://www.slf4j.org/codes.html#no_static_mdc_binder for further details.
```

The solution that they referred was adding this dependency, so I did.

engine/src/main/java/io/camunda/zeebe/process/test/engine/EngineFactory.java
```diff
@@ -24,7 +25,7 @@
   public static InMemoryEngine create() {
     final int partitionId = 1;
     final int partitionCount = 1;
-    final int port = SocketUtil.getNextAddress().getPort();
+    final int port = 26500;
```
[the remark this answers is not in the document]
remcowesterhoud: This port gets used for the grpc server in the container. It used to be this value before we decided maybe we should make it dynamic for parallel testing. The problem when this is dynamic is that I wouldn't know what port it runs. This is an issue when I want to run the container and bind my container port to the internal port. By making it a static value I know how to bind the port.

I don't think it will cause any issues. This port is used inside of the container and since we control the container it feels like a safe assumption that nothing else will be running on this port. We could probably use an environment variable or something to make it configurable, but I wouldn't want to do this right now.

remcowesterhoud:
@korthout Thanks for you review 🙇 I've answered your questions and made some changes. Please check again when you have the time!

pom.xml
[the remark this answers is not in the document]
korthout: Ah I see. 

🔧 In that case, please pull it out into a separate commit explaining why it's needed to accommodate git archeology.

.github/workflows/deploy-artifact.yml
korthout: Interesting. I hadn't thought of that. Thanks for sharing

korthout (approved):
Thanks for the quick changes 🎖️ LGTM

PR #179, opened 2022-02-10
Moved test classes to test sources

remcowesterhoud:
## Description

Somehow these tests ended up in the source files, instead of the test source files. Presumably due to the module split. This moves them back where they belong.

Also modified the junit dependency scope to test. Now that these files are moved the junit is no longer used outside of the test scope.

## Definition of Done

Code changes:

Testing:

Documentation:


saig0 (approved):
LGTM :+1:

PR #190, opened 2022-02-16
Create a filters module

remcowesterhoud:
## Description

A new module "filters" has been introduced. This module will be responsible for filtering the record stream to find specific records.
It has been extracted because this is a shared functionality, regardless of where the record stream comes from. In the future this might move to Zeebe as we have a similar functionality in that project which we could reuse.

## Related issues

Relates to #7. In order to prevent duplication we need this filter logic in a reusable place.

## Definition of Done

Code changes:

Testing:

Documentation:


saig0 (approved):
LGTM :rocket:

PR #191, opened 2022-02-17
Create a test control api

### LINKED ISSUE #7: Add a test control API to the engine
**Description**

Once we have the engine running in a testcontainer we need a way to communicate with the engine. This API should make it possible to:

- Time travel
- Wait for idle state
- Extraction of records

For this API we'd like to gRPC. With gRPC we can define the interface for communicating with the engine. This way the user does not have to bother with mapping requests, they can just generate the interface in their client.

**Related Issues**
#5 
Depends on #24

remcowesterhoud:
## Description

This PR adds a way for the extension to communicate with the containerised engine. A new `engine-protocol` module has been created. In this module the protobuf file will live that defines the interface of communication between the extension and the agent.

This PR also adds the serverside implementation of the protocol. This implementation will live in the `engine-agent`. This is because this agent is responsible for wrapping the engine in a testcontainer. Communication with the engine comes with this responsibility.

## Related issues

closes #7 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/engine-agent/src/main/java/io/camunda/zeebe/process/test/engine/agent/EngineControlImpl.java
+++ b/engine-agent/src/main/java/io/camunda/zeebe/process/test/engine/agent/EngineControlImpl.java
@@ -1,0 +1,106 @@
+package io.camunda.zeebe.process.test.engine.agent;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.camunda.zeebe.process.test.api.InMemoryEngine;
+import io.camunda.zeebe.process.test.engine.EngineFactory;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlGrpc.EngineControlImplBase;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.GetRecordsRequest;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.IncreaseTimeRequest;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.IncreaseTimeResponse;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.RecordResponse;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.ResetEngineRequest;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.ResetEngineResponse;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.StartEngineRequest;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.StartEngineResponse;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.StopEngineRequest;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.StopEngineResponse;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.WaitForIdleStateRequest;
+import io.camunda.zeebe.process.test.engine.protocol.EngineControlOuterClass.WaitForIdleStateResponse;
+import io.camunda.zeebe.protocol.record.Record;
+import io.grpc.Status;
+import io.grpc.stub.StreamObserver;
+import java.time.Duration;
+
+public class EngineControlImpl extends EngineControlImplBase {
+
+  private InMemoryEngine engine;
+
+  public EngineControlImpl(final InMemoryEngine engine) {
+    this.engine = engine;
+  }
+
+  @Override
+  public void startEngine(
+      final StartEngineRequest request,
+      final StreamObserver<StartEngineResponse> responseObserver) {
+    engine.start();
+
+    final StartEngineResponse response = StartEngineResponse.newBuilder().build();
+    responseObserver.onNext(response);
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void stopEngine(
+      final StopEngineRequest request, final StreamObserver<StopEngineResponse> responseObserver) {
+    engine.stop();
+
+    final StopEngineResponse response = StopEngineResponse.newBuilder().build();
+    responseObserver.onNext(response);
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void resetEngine(
+      final ResetEngineRequest request,
+      final StreamObserver<ResetEngineResponse> responseObserver) {
+    engine.stop();
+    engine = EngineFactory.create();
+    engine.start();
+
+    final ResetEngineResponse response = ResetEngineResponse.newBuilder().build();
+    responseObserver.onNext(response);
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void increaseTime(
+      final IncreaseTimeRequest request,
+      final StreamObserver<IncreaseTimeResponse> responseObserver) {
+    engine.increaseTime(Duration.ofMillis(request.getMilliseconds()));
+    final IncreaseTimeResponse response = IncreaseTimeResponse.newBuilder().build();
+    responseObserver.onNext(response);
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void waitForIdleState(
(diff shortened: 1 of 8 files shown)


engine-protocol/src/main/proto/engine_control.proto
```diff
@@ -0,0 +1,64 @@
...
+
+message StartEngineResponse {}
+
+message StopEngineRequest {}
+
+message StopEngineResponse {}
+
+message ResetEngineRequest {}
+
+message ResetEngineResponse {}
```
korthout, line 17: ❓ Did you consider taking the approach we use in Zeebe, i.e. define a common command message with a type property? If so, why this approach over the other?

engine-protocol/src/main/proto/engine_control.proto
```diff
@@ -0,0 +1,64 @@
...
+  // the amount of milliseconds the engine should increase in time
+  int32 milliseconds = 1;
+}
+
+message IncreaseTimeResponse {}
+
+message WaitForIdleStateRequest {
+  // timeout (in ms). The request will be closed if an idle state has not been
+  // achieved withing the timeout.
+  int64 timeout = 1;
```
korthout, line 29: ❓ Do we only want to have a timeout on this specific message?

engine-agent/src/main/java/io/camunda/zeebe/process/test/engine/agent/EngineControlImpl.java
```diff
@@ -0,0 +1,106 @@
...
+    final StopEngineResponse response = StopEngineResponse.newBuilder().build();
+    responseObserver.onNext(response);
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void resetEngine(
+      final ResetEngineRequest request,
+      final StreamObserver<ResetEngineResponse> responseObserver) {
+    engine.stop();
```
korthout, line 58: ❓ Can stopping the engine fail? If so, what happens?

korthout (changes requested):
Sorry for the late review @remcowesterhoud 

👍 I like the general direction, and thanks for the good explanation in the commit messages and comments. Great job! 

However, I'm not that familiar with some of these topics. So, I still have a bunch of questions that might sound a bit dumb or redundant. Please bear with me. It would be nice if you could clarify them.

❌ The main thing that's missing is tests. Please make sure to add some.

engine-protocol/src/main/proto/engine_control.proto
remcowesterhoud: Yes. This is the only request where we will be waiting for something on the server-side. We don't want to get stuck so a timeout here seems like a good addition. This timeout is not the request timeout, it's the waiting timeout. gRPC can handle request timeouts itself.

engine-protocol/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+        </configuration>
+        <executions>
+          <execution>
+            <goals>
+              <goal>compile</goal>
+              <goal>compile-custom</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
```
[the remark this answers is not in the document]
remcowesterhoud: This plugin is for the code generation. This is how the gRPC docs recommend doing this https://grpc.io/docs/languages/java/generated-code/#codegen

pom.xml
```diff
@@ -325,7 +351,15 @@
...
       </plugin>
+
     </plugins>
+    <extensions>
+      <extension>
+        <groupId>kr.motd.maven</groupId>
+        <artifactId>os-maven-plugin</artifactId>
+        <version>${plugin.version.os-maven}</version>
+      </extension>
+    </extensions>
```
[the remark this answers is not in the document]
remcowesterhoud: The `os-maven-plugin` generates various useful platform-dependent project properties. We need this for the `protobuf-maven-plugin` for the `${os.detected.classifier}` property.

engine-protocol/src/main/proto/engine_control.proto
remcowesterhoud: Not sure what you mean here. In Zeebe we do the same thing as I did here looking at `gateway.proto`. Care to elaborate?

engine-agent/src/main/java/io/camunda/zeebe/process/test/engine/agent/EngineControlImpl.java
remcowesterhoud: It could fail. If it happens an error will be thrown which will cause in a failed gRPC request. In the end it would result in a failed unit test. I've added logging in case this happens so we can see what went wrong. I've never had it actually happen before.

remcowesterhoud:
@korthout Thanks for reviewing! I've answered your questions and made some changes. Please have another look.

About the unit tests. This will all be extensively tested by the QA tests once the client-side implementation gets merged and this code actually gets used. I wanted to keep the PR's a bit more small and isolated.

engine-protocol/src/main/proto/engine_control.proto
korthout: 👍 Oops. I seem to have mixed-up the gateway protocol (gRPC) and the broker's protocol (SBE). 

😅 I guess it was Friday afternoon

engine-agent/src/main/java/io/camunda/zeebe/process/test/engine/agent/EngineControlImpl.java
```diff
@@ -96,7 +100,7 @@ public void getRecords(
             RecordResponse.newBuilder().setRecordJson(recordJson).build();
         responseObserver.onNext(response);
       } catch (JsonProcessingException e) {
-        e.printStackTrace();
+        LOG.error("Failed mapping record to JSON", e);
```
korthout, line 103: 👍 Thanks this looks better

🔧 I'm not sure what information will be available in JsonProcessingException. It might be useful to add some information about the specific record that couldn't be mapped. E.g. `LOG.error("Failed to map record {} at position {} on partition {} to JSON", record.getKey(), record.getPosition(), record.getPartitionId(), e)`

engine-protocol/src/main/proto/engine_control.proto
```diff
@@ -34,15 +34,27 @@ message WaitForIdleStateResponse {}
...
+    Stops the in-memory engine.
+   */
   rpc StopEngine (StopEngineRequest) returns (StopEngineResponse);
 
+  /*
+    Resets the in-memory engine. This is done by recreating the engine entirely.
+    The engine should be reset before executing the next test. This prevents
+    the tests from influencing each other.
+   */
   rpc ResetEngine (ResetEngineRequest) returns (ResetEngineResponse);
```
korthout, line 58: 👍 I like it!

🔧 I seems to me that there is still a chance for tests to influence each other. If 2 tests run concurrently, and 1 resets the engine, it affects the other test. Right? If so, please make it clear from this doc that the user should still watch out for such situations.

engine-protocol/src/main/proto/engine_control.proto
```diff
@@ -58,7 +70,10 @@ service EngineControl {
    */
   rpc WaitForIdleState (WaitForIdleStateRequest) returns (WaitForIdleStateResponse);
 
-
+  /*
+    Get all records from the in-memory engine. These records will be returned
+    in a JSON format. Client-side these should be mapped to Records.
+   */
```
korthout, line 76: ❓ Should this documentation refer to the specific JSON-schema used for these records? I'm not sure, what do you think?

korthout (approved):
Thanks for the changes @remcowesterhoud LGTM 🚀

engine-protocol/src/main/proto/engine_control.proto
remcowesterhoud: Do we have these schemas available somewhere already? I think only a very select few users would ever implement the gRPC client-side themselves, usually it's handled by the extension. But it can't hurt to add it I suppose.

korthout: I'm not really sure what the schema is here. But, I guess it refers to https://github.com/camunda-cloud/zeebe/tree/main/protocol. If it's about deserialization, we can refer to https://github.com/camunda-cloud/zeebe/tree/main/protocol-jackson as an easy way to achieve the mapping.

Issue #192, opened 2022-02-17
Use testcontainers in the extension


remcowesterhoud:
**Description**

To support java8 the extension should start making use of testcontainers. With #5 and #7 the preparation for this has been done.
What still needs to be done:

1. Start the testcontainer in the extension
2. Implement the client side of the `engine-protocol` so we can communicate with the testcontainer.

PR #193, opened 2022-02-17
Isolate shared RecordStreamSource functionality

remcowesterhoud:
## Description

The `RecordStreamSource` interface defines a lot of methods that generic and should be reusable. This PR creates a new `RecordStream` class in the `filters` module. This class will implement these generic functions, so not all `RecordStreamSource` classes have to implement them separately, preventing duplication.

It's quite a refactoring so I've done it in a few steps to make it easier to review:
1. Create the `RecordStream` class containing the shared functionality.
2. Remove the shared functionality from the `RecordStreamSource` interface and the implementation.
3. Use the new `RecordStream` class in the `engine` module.
4. Make the filters in the `filters` module use the new `RecordStream` class.
5. Switch the `RecordStreamSource` to the new `RecordStream` in the other modules.

In Zeebe we have a similar functionality in the `RecordingExporter`. It would be nice if we could share this functionality between the 2 projects at some time. We can't use the zeebe implementation at the moment because it is not Java 8 compatible.

## Related issues

This is a requirement for #192 . Without doing this #192 would introduce a lot of duplication as it needs create an implementation of the `RecordStreamSource` interface.

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/filters/src/main/java/io/camunda/zeebe/process/test/filters/RecordStream.java
+++ b/filters/src/main/java/io/camunda/zeebe/process/test/filters/RecordStream.java
@@ -1,0 +1,128 @@
+package io.camunda.zeebe.process.test.filters;
+
+import io.camunda.zeebe.process.test.api.RecordStreamSource;
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.RecordValue;
+import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.protocol.record.value.DeploymentRecordValue;
+import io.camunda.zeebe.protocol.record.value.IncidentRecordValue;
+import io.camunda.zeebe.protocol.record.value.JobBatchRecordValue;
+import io.camunda.zeebe.protocol.record.value.JobRecordValue;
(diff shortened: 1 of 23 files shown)


filters/src/main/java/io/camunda/zeebe/process/test/filters/RecordStream.java
```diff
@@ -0,0 +1,128 @@
...
+
+public class RecordStream {
+
+  private static final Logger LOG = LoggerFactory.getLogger("io.camunda.zeebe.process.test");
+
+  private RecordStreamSource recordStreamSource;
+
+  public RecordStream(final RecordStreamSource recordStreamSource) {
+    this.recordStreamSource = recordStreamSource;
+  }
```
saig0, line 35: :wrench: We could change the visibility of the constructor to `private`. 

`RecordStream` has a static factory method `of(...)` that is used in the code. The constructor is not used except in the factory method.

saig0 (approved):
@remcowesterhoud looks good :rocket: 

I have just a small suggestion. Please have a look before merging.

PR #207, opened 2022-02-24
Improve failure logging

### LINKED ISSUE #188: Inform developer about incidents happening in a test case (possibly marking invalid expressions)
When I run a process test, I might encounter situations I have not foreseen, this is exactly the nature of writing tests :-)

I used zeebe-process-test 1.3.3 to test this process and actually used an invalid expression in the user task assignment. I can see this clearly in Operate if I run the process manually:

grafik
grafik

However, in my test case, I simply see, that I do not "arrive" in the user task, without any information about what just happened. 

So in my case, I did some manual exploration to finally see the problem, which is now easy to fix. But this was an annoying detour.

I think we need to guide the user better than this. Spontaneous ideas:

- Log a warning every time an incident is created? I know that there might be situations where you want to test the incident, but I consider this rare - and the bigger problem is actually incidents going unnoticed, like in my example.
- Probably introduce some method to verify that there are no incidents. But I actually would assume that people do not know about it and forget to call it. Probably `inMemoryEngine.waitForIdleState()` should throw an exception if it ends with an incident?
- ?

remcowesterhoud:
## Description

It's important users are able to debug their process properly if a test fails. To enable this we should provide them with proper logging upon a failure.
In the past we used the CompactRecordLogger from zeebe-test-util for this. This solution had 2 problems:
1. zeebe-test-util is not bound to Java 8. This means users running on Java 8 cannot use it.
2. The CompactRecordLogger provides too much information for the targeted users of this project.

To solve these problems I have introduced a new logger. This will also log all records of the stream, but only logs the most important fields that users need for debugging their processes. At the moment it is a guess which fields these are. Once we receive user feedback we can easily modify the logging for the specific records that are not clear enough.

## Related issues

closes #188 

## Definition of Done

Code changes:

Testing:

Documentation:


korthout (changes requested):
I really like these changes @remcowesterhoud Great job! 👍 

❌ I have a few improvements we should consider before merging. Mainly, I want to avoid confusing users unnecessarily.

filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
```diff
@@ -0,0 +1,291 @@
...
+        ValueType.MESSAGE_START_EVENT_SUBSCRIPTION, this::logMessageStartEventSubscriptionRecord);
+    valueTypeLoggers.put(ValueType.VARIABLE, this::logVariableRecord);
+    valueTypeLoggers.put(ValueType.VARIABLE_DOCUMENT, this::logVariableDocumentRecord);
+    valueTypeLoggers.put(
+        ValueType.PROCESS_INSTANCE_CREATION, this::logProcessInstanceCreationRecord);
+    valueTypeLoggers.put(ValueType.ERROR, this::logErrorRecord);
+    valueTypeLoggers.put(ValueType.PROCESS_INSTANCE_RESULT, this::logProcessInstanceResultRecord);
+    valueTypeLoggers.put(ValueType.PROCESS, this::logProcessRecord);
+    valueTypeLoggers.put(ValueType.PROCESS_EVENT, this::logProcessEventRecord);
+  }
```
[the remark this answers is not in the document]
remcowesterhoud: I understand why you'd say this. I didn't include all value types because these value types don't provide any extra information that would be useful for a user.

I've added the test but I do think it's gonna be an annoying situation for release managers. After all that is the moment where we will update the Zeebe version and encounter the failing tests.

extension/src/main/java/io/camunda/zeebe/process/test/extension/ZeebeProcessTestExtension.java
```diff
@@ -61,7 +64,7 @@ public void testFailed(final ExtensionContext extensionContext, final Throwable
     final Object engineContent = getStore(extensionContext).get(KEY_ZEEBE_ENGINE);
     final InMemoryEngine engine = (InMemoryEngine) engineContent;
 
-    System.out.println("===== Test failed! Printing records from the stream:");
+    LOG.error("===== Test failed! Printing records from the stream:");
```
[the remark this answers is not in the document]
remcowesterhoud: I think it's a great idea 👍

filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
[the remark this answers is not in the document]
korthout: >it's gonna be an annoying situation for release managers

Yes, that is the cost of this approach. The alternative is that users might see the generic record logger's output + us not being aware of it.

With the test in place, the worst case is that we have to temporarily ignore the test to release a version of zeebe-process-test. Right?

remcowesterhoud:
@korthout Thanks a lot for all the reviews you're doing for me! 🏆 

You raised some valid points here and I've made the necessary changes. Please have another look 🙇

filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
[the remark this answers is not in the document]
remcowesterhoud: Yea, either that or just adding the value type logger which isn't that much work either

filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
```diff
@@ -77,18 +78,26 @@ public void log() {
...
           .append("Incidents occurred:")
           .append(System.lineSeparator());
-      incidents.forEach(
-          record -> stringBuilder.append(summarizeIncident(record)).append(System.lineSeparator()));
+      createIncidents.forEach(
+          record -> {
+            if (!resolvedIncidents.contains(record.getKey())) {
+              stringBuilder.append(summarizeIncident(record)).append(System.lineSeparator());
+            }
+          });
```
korthout: 🔧  You can filter the createIncidents first and then run forEach:

```java
// pseudocode
createIncidents
  .filter(not(record -> resolvedIncidents.contains(record.getKey()))
  .forEach(record -> stringBuilder.append(summarizeIncident(record)).append(System.lineSeparator()));
```

filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
```diff
@@ -169,8 +178,10 @@ private String logProcessInstanceRecord(final Record<?> record) {
   private String logIncidentRecord(final Record<?> record) {
     final IncidentRecordValue value = (IncidentRecordValue) record.getValue();
     final StringJoiner joiner = new StringJoiner(", ", "", "");
-    joiner.add(String.format("(Element id: %s)", value.getElementId()));
-    joiner.add(String.format("(Process id: %s)", value.getBpmnProcessId()));
+    if (record.getRecordType().equals(RecordType.EVENT)) {
```
korthout, line 132: ❌ We should also log commands and command rejections. Unless we decided to only print events

filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
```diff
@@ -138,9 +138,9 @@ private void logRecords(final StringBuilder stringBuilder) {
 
   private String logGenericRecord(final Record<?> record) {
     return System.lineSeparator()
-        + String.format("| %-15s", record.getRecordType())
-        + String.format("%-30s", record.getValueType())
-        + String.format("%-25s| ", record.getIntent());
+        + String.format("| %-20s", record.getRecordType())
+        + String.format("%-35s", record.getValueType())
+        + String.format("%-30s| ", record.getIntent());
```
korthout, line 94: 💭 I worry this becomes too long for many users, but we can leave it for now.

korthout (approved):
Nice work @remcowesterhoud 🦖 

🔑 I have one last change request, but I already want to provide my approval so this isn't blocked by me anymore. So here's a key to represent me handing this over 😆

filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
remcowesterhoud: We always print every Record. For some commands however the details are empty. For example for this incident, if we receive a command no `elementId` or `bpmnProcessId` will be set.

korthout: 🔧 Aha, in that case, let's just base it on the properties themselves. If the property has a value, let's print it.

🔧 It might help to rename these methods to `logIncidentRecordValue`, to differentiate it from logging the Record, WDYT?

PR #209, opened 2022-02-24
Flatten all POMs for deployment

### LINKED ISSUE #208: Introduce flatten pom
# Description

We should introduce the flatten-maven-plugin, which will flatten any hierarchy in a module's POM before it's deployed somewhere. The original POM will remain intact. The goal here is to simplify the integration of the modules in other people's projects, by making it more explicit which dependencies are pulled by the module, and which version they should be.

remcowesterhoud:
## Description

Introduces the `flatten-maven-plugin` in the build process. The projects are flattened according to the `ossrh` rules.

## Related issues

closes #208

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,6 +15,11 @@
 
   <name>Zeebe Process Test Root</name>
 
+  <description>
+    This pom should be the parent of all specified modules. It defines all versions and provides
+    plugins that are required for the child modules.
+  </description>
+
   <modules>
     <module>api</module>
     <module>assertions</module>
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,7 @@
     <nexus.snapshot.repository>https://app.camunda.com/nexus/content/repositories/zeebe-io-snapshots</nexus.snapshot.repository>
     <nexus.snapshot.repository.id>camunda-nexus</nexus.snapshot.repository.id>
 
+    <plugin.version.flatten>1.2.2</plugin.version.flatten>
     <plugin.version.fmt>2.9.1</plugin.version.fmt>
     <plugin.version.googlejavaformat>1.12.0</plugin.version.googlejavaformat>
     <plugin.version.jacoco>0.8.7</plugin.version.jacoco>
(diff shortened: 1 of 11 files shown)


pom.xml
```diff
@@ -377,6 +378,41 @@
         </executions>
       </plugin>
 
+      <!-- Flatten pom plugin -->
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>flatten-maven-plugin</artifactId>
+        <version>${plugin.version.flatten}</version>
+        <configuration combine.children="append">
```
korthout, line 386: 🔧 The `combine.children` is not necessary here, because this configuration is not overwriting or adding to an inherited configuration for the flatten plugin. It's only defined here. The attribute can be removed.

pom.xml
```diff
@@ -377,6 +378,41 @@
...
+          <flattenMode>ossrh</flattenMode>
```
korthout, line 387: 🔧 The `ossrh` mode of the flatten plugin removes unnecessary tags and inlines properties and relationships. It highlights some inherited tags from the parent of this pom `camunda-release-parent`. Please overwrite the `description`:
- in the root pom as empty
- or, in the root pom as a description of the root pom + in all other poms with a description of that project.

💭 Also note that the `developers` section is currently set to:

```xml
  <developers>
    <developer>
      <id>camunda.org</id>
      <name>camunda BPM Community</name>
      <organization>camunda.org</organization>
      <organizationUrl>http://camunda.org</organizationUrl>
    </developer>
  </developers>
```

This is the same as zeebe, but I'm not sure whether it's actually correct there. So I'm also not sure about it for this project. Please give it some thought.

korthout (approved):
Looking good @remcowesterhoud

I have a few suggestions, but nothing blocking 👍

remcowesterhoud:
@korthout Thanks for the review 🙇 

I've added descriptions to all modules and removed the `append.children` property like you suggested.

PR #210, opened 2022-02-24
Remove public from test methods

### LINKED ISSUE #197: Remove public keyword from test methods
**Description**

We'd like to keep the scope of methods as small as possible. Even though these are tests it'd be nice to keep it in line with the rest of the code. JUnit 5 allows us to get rid of the `public` keyword. In some tests we don't have them, but in others we still do. We should get rid of all of them.

remcowesterhoud:
## Description

This PR removes the `public` keywords from our test methods.
jUnit5 supports tests methods without a public keyword. By convention we want methods to be scope as small as possible. We should also adhere to this in our test classes.

## Related issues

closes #197 

## Definition of Done

Code changes:

Testing:

Documentation:


saig0 (approved):
LGTM :rocket:

PR #211, opened 2022-02-24
Set up licensing

### LINKED ISSUE #171: Set up licensing
**Description**

We need to properly license our code. None of this project is part of the core-business. Therefore the project can be licensed under the regular Apache 2 license.
One exception is the in memory engine. This module depends on the zeebe broker, which is part of the core-business. This means that the engine module must follow the same license as the zeebe broker.

**What we need to do:**
1. Create a license folder like in zeebe (https://github.com/camunda-cloud/zeebe/tree/main/licenses) containing the apache license and the zeebe license.
2. Add a section to the README and the Contribution Guide specifying that all modules are licensed under Apache 2, except for the engine which is licensed under the Zeebe license.
3. Remove the LICENSE file

remcowesterhoud:
## Description

This PR set ups the licensing for this project. None of this project is part of the core-business of Camunda. Therefore the project can be licensed under the regular Apache 2 license.

There are 3 exceptions. These are the modules that depend on the zeebe broker, which is part of the core-business. Because of this these modules have to follow the same license as the broker, which is the Zeebe Community License V1.1. These modules are:
1. engine
2. engine-agent
3. extension

I've also set IntelliJ up to automatically add the correct copyright headers when a new class is created.
 
## Related issues

closes #171 

## Definition of Done

Code changes:

Testing:

Documentation:


remcowesterhoud:
@pihme I recommend reviewing by commit, as copyright headers were added to every java file we have.

licenses/ZEEBE-COPYING-HEADER.txt
```diff
@@ -0,0 +1,5 @@
+Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+one or more contributor license agreements. See the NOTICE file distributed
```
pihme, line 2: :question: What is the NOTICE file? Is this generated by Maven and shipped in the JAR or should this live in the repo?

engine-agent/pom.xml
```diff
@@ -18,6 +18,13 @@
   <description>A wrapper around the test engine. This wrapper makes it possible to run the engine in a
     testcontainer.</description>
 
+  <licenses>
+    <license>
+      <name>Zeebe Community License v1.1</name>
+      <url>https://zeebe.io/legal/zeebe-community-license-1.1</url>
```
pihme, line 24: :question: Not sure what the legal best practice is here. Should we link to the source or link to the copy in this repo?

.idea/copyright/zeebe_copyright.xml
```diff
@@ -0,0 +1,6 @@
+<component name="CopyrightManager">
+  <copyright>
+    <option name="notice" value="Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under&#10;one or more contributor license agreements. See the NOTICE file distributed&#10;with this work for additional information regarding copyright ownership.&#10;Licensed under the Zeebe Community License 1.1. You may not use this file&#10;except in compliance with the Zeebe Community License 1.1.&#10;" />
```
pihme, line 3: :question: Again `NOTICE` file

.idea/copyright/zeebe_copyright.xml
```diff
@@ -0,0 +1,6 @@
...
+    <option name="myName" value="zeebe-copyright" />
```
pihme, line 4: :wrench: Maybe rename file and shorthand to _Zeebe Community License_  which seems more precise

pihme (approved):
Several questions. Most notably I wonder that the `NOTICE` file is referring to.

If you make changes in this PR, I would like to see it again. 

If everything is sound, feel free to merge it right away.

pihme:
One thing that I wonder is whether we should also have a clarification like this: https://github.com/camunda-cloud/zeebe#clarification-on-grpc-code-generation

This repo also defines a GRPC protocol. And if anybody wanted to use it with a different language, they could use the InMemory engine in the container and would need to write their own way to interact with the container. So in that context, I assume the same clarification would apply.

remcowesterhoud:
I think a big difference in our case is that our protocol is licensed with Apache 2, whereas the gateway-protocol is licensed with Zeebe Community 1.1.

The InMemoryEngine is also licensed with Apache 2, so anyone that wants to implement this to use it with a different language is free to do so. They wouldn't have any need to do anything with the actual implementation of the InMemoryEngine.

licenses/ZEEBE-COPYING-HEADER.txt
remcowesterhoud: Good question. It seems like something I'd have to add to the repo. I'm not quite sure what this is supposed to contain though. The one in Zeebe says:
```
Camunda
Copyright 2017-2019 Camunda Services GmbH
```
Which seems a bit outdated 😅

engine-agent/pom.xml
remcowesterhoud: I just did what we did in Zeebe here and assumed it is the way to go.

.idea/copyright/zeebe_copyright.xml
remcowesterhoud: If by shorthand you mean `myName` then unfortunately there is no option to change it. I can rename it manually, but IntelliJ will override it when any changes are made to the copyright settings.

remcowesterhoud:
@pihme I've added some minor changes. Please have another look 🙂

pihme (approved):
looks good to me

PR #218, opened 2022-02-28
Add waitForProcessingState to engine

### LINKED ISSUE #202: Flaky test: ProcessEventInspectionsTest.testFindLastProcessInstance
**Description**

It seems to be a timing issue. This process contains a timer and the test increases the time of the engine: `Utilities.increaseTime(engine, Duration.ofDays(1));`. Possibly the engine has not yet triggered the timer and started processing before we checked if it's idle. This means we consider the engine to be idle, even though it still has stuff to do (triggering the timer).

```
java.lang.AssertionError: 

Expecting Optional to contain a value but it was empty.
	at io.camunda.zeebe.process.test.qa.inspections.ProcessEventInspectionsTest.testFindLastProcessInstance(ProcessEventInspectionsTest.java:61)
```

Occurs with both extensions.

remcowesterhoud:
## Description

In #202 we have a flaky tests. In this test we have a process that's triggered by a timer start event. We increase the time of the engine to trigger this timer, sleep for 100 ms, wait for the engine to reach an idle state and assert that the process instance gets completed.

This solution has a few problems. First of all we use a `Thread.sleep()` to give the engine some time to trigger the timers. This is a dirty solution and it'd be better if we have a solution that only waits until the engine is running again.
The second problem is that the engine has not always triggered the timers within 100 ms. Because of this the engine is considered idle, without actually finishing the processing.

This PR adds a `waitForProcessingState` to the engine. This method is very similar to the `waitForIdleState` method we already have. Just like that method it uses the `IdleStateMonitor` with a callback to wait until the engine has started committing records again.
A timeout needs to be specified to prevent waiting forever in case the engine will be permanently idle.

## Related issues

closes #202 

## Definition of Done

Code changes:

Testing:

Documentation:


api/src/main/java/io/camunda/zeebe/process/test/api/InMemoryEngine.java
```diff
@@ -29,10 +30,15 @@
...
-   * Waits for the engine to reach an idle state.
+   * Waits for the engine to start processing.
    *
-   * <p>For more info on the idle state refer to {@link IdleStateMonitor}
+   * @param timeout the maximum amount of time to wait before processing state has been reached
+   * @throws InterruptedException if the current thread was interrupted while waiting
+   * @throws TimeoutException if the engine has not reached a processing state before the timeout
    */
-  void waitForIdleState();
+  void waitForProcessingState(Duration timeout) throws InterruptedException, TimeoutException;
```
pihme, line 43: :wrench: 
```suggestion
  void waitForBusyState(Duration timeout) throws InterruptedException, TimeoutException;
```

Maybe also change order of methods:
1. Wait for Busy
2. Wait for Idle

pihme: :wrench: Maybe harmonize interfaces and also add a timeout to `waitForIdleState`

engine/src/main/java/io/camunda/zeebe/process/test/engine/IdleStateMonitor.java
```diff
@@ -44,9 +45,16 @@
...
   }
 
-  public void addCallback(final Runnable callback) {
-    synchronized (callbacks) {
-      callbacks.add(callback);
+  public void addOnIdleCallback(final Runnable callback) {
+    synchronized (idleCallbacks) {
+      idleCallbacks.add(callback);
+    }
+    checkIdleState();
```
pihme, line 52: :x: rename method to something like `checkEngineState` or `checkProcessingOrIdleState` or something along those lines.

Maybe we should also rename the method to `checkStateAndNotifyCallbacks` to hint to the side effect

qa/src/test/java/io/camunda/zeebe/process/test/qa/util/Utilities.java
```diff
@@ -149,15 +150,16 @@ public static PublishMessageResponse sendMessage(
...
-  public static void increaseTime(final InMemoryEngine engine, final Duration duration) {
+  public static void increaseTime(final InMemoryEngine engine, final Duration duration)
+      throws InterruptedException {
     engine.increaseTime(duration);
     try {
-      // we need to wait some physical time so that InMemoryEngine has a chance to fire the timers
-      Thread.sleep(100);
-    } catch (InterruptedException e) {
-      // do nothing
+      engine.waitForProcessingState(Duration.ofSeconds(1));
```
pihme, line 157: :wrench: maybe also add a `waitForProcessingState` method to this class so the two lines look more alike

api/src/main/java/io/camunda/zeebe/process/test/api/InMemoryEngine.java
pihme: btw, totally fine to keep the name processing state in the idle state monitor, because there it is more technical and closer to the implementation. But here I have a slight preference for using the busy idle pair. But it's only a mild preference, so I am also ok with sticking with the name.

engine/src/main/java/io/camunda/zeebe/process/test/engine/IdleStateMonitor.java
pihme: :x: dito for renaming the class altogether

pihme (changes requested):
Looks good!

Some suggestions for name changes.

And maybe we should start writing tests also for classes like the idle state. The longer we wait, the more painful it will be to start writing tests for a class like that.

remcowesterhoud:
Thanks for reviewing @pihme. I've applied your suggestions and created a unit test for the `IdleStateMonitor`. The test is somewhat complex because I needed to make my own `LogStreamReader`. If you see a better solution for this please let me know!

engine/src/main/java/io/camunda/zeebe/process/test/engine/EngineStateMonitor.java
```diff
@@ -49,22 +50,22 @@ public void addOnIdleCallback(final Runnable callback) {
...
-  private void checkIdleState() {
+  private void checkEngineStateAndNotifyCallbacks() {
     synchronized (idleStateNotifier) {
       if (isInIdleState()) {
-        scheduleNotification();
+        scheduleIdleStateNotification();
       } else {
-        cancelNotification();
+        cancelIdleStateNotification();
         runProcessingCallbacks();
```
pihme, line 69: 🔧 maybe rename to `notifyProcessingCallbacks`

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineStateMonitorTest.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import org.junit.jupiter.api.Test;
+
+class EngineStateMonitorTest {
+
+  @Test
+  void testCallbackIsCalledWhenEngineIsAlreadyIdle()
```
pihme, line 18: :x: please qualify which callback is being called in test methods names (dito for the following)

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineStateMonitorTest.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  @Test
+  void testCallbackIsCalledWhenIdleStateIsReached()
+      throws ExecutionException, InterruptedException, TimeoutException {
+    // given
+    final InMemoryLogStorage logStorage = mock(InMemoryLogStorage.class);
+    final TestLogStreamReader logStreamReader = new TestLogStreamReader();
+    final EngineStateMonitor monitor = new EngineStateMonitor(logStorage, logStreamReader);
+    changeToBusyState(monitor, logStreamReader);
```
pihme, line 42: 🤔 I wonder what would be clearer to read. We do the setup of the callbacks in _given_ and then in _when_ we change to busy state. Or we do it like it is now. But, personally, I find this hard to read. Because the when/then part is _when I register a callback, it is being called_. 

I like this test implementation, because it has an explicit call to _changeToBusyState_ I wonder whether we could also manufacture a _changeToIdleState_ method alongside and use it in the test case above.

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineStateMonitorTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+  private void changeToBusyState(final EngineStateMonitor monitor, final TestLogStreamReader reader) {
+    // We use onReplayed here because it is an easy way to update the lastProcessedEventPosition of
+    // the EngineStateMonitor
+    final long position = reader.getLastEventPosition() + 1;
+    monitor.onReplayed(position, -1L);
+    reader.setLastEventPosition(position);
+  }
+
+  private class TestLogStreamReader implements LogStreamReader {
```
pihme, line 114: :wrench: I think you could just use mocks for this. But I don't think this too verbose, either.

pihme (approved):
Looks good!

I have some comments. 

If you make bigger changes to the test, I would like to review those again. Otherwise, no need to do another review round

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineStateMonitorTest.java
pihme: Ah, found the _changeToIdleState_ method.

remcowesterhoud: It's hard to structure the given/when/then like you suggested. This is because when we add a callback to the `EngineStateMonitor` it immediately checks if the engine is busy or idle, and then notifies the callbacks if necessary. To circumvent this we need to call `changeToBusyState` or `changeToIdleState` before we actually add the callback.

I will make some other changes to make it more readable and less confusing though.

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineStateMonitorTest.java
remcowesterhoud: I'd rather stick with this solution in that case. I think mocking is gonna make the test more confusing.

remcowesterhoud:
@pihme I made some bigger changes to the test so I'd like another review.

The biggest change is that I added a way to lock the state. This is required because when we switch to a busy state and register a callback the check if we are already in an idle state. This executes this method:

```java
private void forwardToLastEvent() {
    synchronized (reader) {
      while (reader.hasNext()) {
        lastEventPosition = reader.next().getPosition();
      }
    }
  }
```

This iteration causes the state to become idle. Before this change as a workaround I registered the busy callback before actually changing the state to busy. This was confusing so I switched this up, but because of it I required a way to lock the state into a busy state. That's why I added the `stateLocked` flag.

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineStateMonitorTest.java
```diff
@@ -161,7 +165,7 @@ public void close() {}
 
     @Override
     public boolean hasNext() {
-      return position < lastEventPosition;
+      return !stateLocked && position < lastEventPosition;
```
pihme, line 168: :x: I think this is a bit misleading in terms of naming and what it does. It seems to lock only one of the states. 

I think it would be clearer to either rename `stateLocked` or to capture the state at the moment it is locked:
````
setStateLocked(stateLocked) {
  this.stateLocked = stateLocked;
  if (stateLocked) {
     capturedState = position < lastPosition; 
  }
}

hasNext()  {
 return stateLocked ? capturedState : position < lastPosition
}
````

**Review by pihme — CHANGES_REQUESTED** [review]
Quick feedback; didn't do a full review yet

File: engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineStateMonitorTest.java [hunk 9]
```diff
@@ -75,57 +76,60 @@ void testCallbackIsCalledWhenEngineIsAlreadyBusy()
   }
 
   @Test
-  void testCallbackIsCalledWhenBusyStateIsReached()
+  void testOnProcessingCallbackIsCalledWhenBusyStateIsReached()
```
pihme, line 79: :x: If I am not mistaken this test now has exactly the same code as `...IsAlreadyBusy`

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineStateMonitorTest.java
```diff
@@ -10,40 +10,48 @@
...
-    final TestLogStreamReader logStreamReader = new TestLogStreamReader();
-    final EngineStateMonitor monitor = new EngineStateMonitor(logStorage, logStreamReader);
-    changeToBusyState(monitor, logStreamReader);
-
-    // when
     final CompletableFuture<Void> callbackFuture = new CompletableFuture<>();
     final Runnable callback = () -> callbackFuture.complete(null);
+
+    // when
+    changeToBusyState(monitor, logStreamReader, false);
```
pihme, line 54: It's not so clear to see which line in the when part is responsible for reaching the idle state.

pihme (changes requested):
Definitely going in the right direction, Some details are still confusing

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineStateMonitorTest.java
remcowesterhoud: I'm not sure what you mean here. `stateLocked` definitely locks both of the states.

If the state is locked `!stateLocked && position < lastEventPosition;` will always resolve as `false`. This has as a result that the `EngineStateMonitor` will never update it's position. This means that if the positions are equal, the state is locked to idle. If the positions are not equal the state is locked to busy.
```java
synchronized (reader) {
      while (reader.hasNext()) { // <--- this will resolve to false so the lastEventPosition is never updated.
        lastEventPosition = reader.next().getPosition();
      }
    }
```

pihme: Hmm, I see what you mean. The thing that is confusing me is that `hasNext` is always false when `stateLocked=true`. Which seems wrong, in my mind. And the only thing it does work is indirectly because of taking into account implementation details of how this method is used as a guard around the call to next.

Maybe another way could be to implement it like this:
```
    @Override
    public LoggedEvent next() {
     if !stateLocked {
        position++;
    }
      final LoggedEvent event = mock(LoggedEvent.class);
      when(event.getPosition()).thenReturn(position);
      return event;
    }
```

So then `hasNext` would give always the correct result, but one couldn't really advance. But maybe this then results in endless loops 🤔

remcowesterhoud: Yes, this solution will result in infinite loops.

I don't like the other solution either though:
```java
setStateLocked(stateLocked) {
  this.stateLocked = stateLocked;
  if (stateLocked) {
     capturedState = position < lastPosition; 
  }
}

hasNext()  {
 return stateLocked ? capturedState : position < lastPosition
}
```
This way it seems like. the state is being managed by the reader which is not the case.

The way it is currently makes sense in my mind. The `EngineStateMonitor` asks if there is any more events it should process. Because the state is locked we override this by saying, no there isn't anything. Thus, keeping the state of the `EngineStateMonitor` the same.

pihme: What about we time box this, each of us thinks 30mins hard about the problem. If we find no better solution, we leave it as it is?

remcowesterhoud: Sounds like a plan

pihme: I think this is goo enough for now

pihme (approved):
🎉

PR #226, opened 2022-03-02
Release/1.4.0 alpha2

remcowesterhoud:
## Description

Several fixes in the workflow to deal with the fact that git is in headless mode when checking out a release tag.

1. The deploy to maven and the deploy to docker jobs have been combined into a single deploy job
2. Maven is now used to set the image tag in `extension-testcontainer` config file.
3. A new step had been introduces to make sure changes on the `main` branch are still using the `latest` tag. This way we keep the same behaviour as we had previously.

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -4,7 +4,7 @@ on:
   workflow_call: { }
 jobs:
   code-formatting:
-    name: Apply code formatting
+    name: check code formatting
     timeout-minutes: 15
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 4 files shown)


remcowesterhoud:
@pihme I've fixed the workflow. I think this covers all cases.

1. Commits on main will create an image with the `latest` tag. SNAPSHOT releases will have `latest` in the config. The `Update tag in config` step which is only run on the `main` branch will make sure maven doesn't override the tag in the config with the `${project.version}`.
2. All releases will create an image with the tag set to the released version. Maven will take care of this version in the config by overriding the `${project.version}`.

pihme (approved):
Looks good to me.

The only concern I have that in two months we will have forgotten how it works exactly. Maybe you could add some comments or documentation to explain the solution

PR #230, opened 2022-03-03
Cleanup compile time dependencies

### LINKED ISSUE #174: Review compile time dependencies
**Description**

Right now, adding the project as a dependency has the downside of including Log4J as the SLF4J implementation. If your project already includes it, then you get conflicts and it's unclear which implementation will get used. So the implementation in this project should be a test scoped dependency.

I think there may also be other dependencies which should not be compile time, but I imagine you have a better than me on that account. Happy to discuss this f2f.

NOTE: I think this may go away once we containerize the engine, but as we may end up supporting embedding the engine anyway, I think it's worth doing regardless.

NOTE: I would also propose using just `slf4-simple` as an implementation if it's only for testing, no need to bring in log4j.

remcowesterhoud:
## Description

This PR gets rid of dependencies to any `slf4j` implementation. Instead we make use of the `slf4j-api`. In our tests we still need to provide an implementation which is done by adding `slf4j-simple` as a test scoped dependency. I got rid of `log4j` as `slf4j-simple` is more lightweight.

I've also gone through the other dependencies. As a result of the module split some of these were unused, or wrongly scoped. This PR also solves these issues.

## Related issues

closes #174 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/assertions/pom.xml
+++ b/assertions/pom.xml
@@ -43,16 +43,6 @@
       <artifactId>assertj-core</artifactId>
     </dependency>
 
-    <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-api</artifactId>
-    </dependency>
-
-    <dependency>
-      <groupId>io.camunda</groupId>
-      <artifactId>zeebe-client-java</artifactId>
-    </dependency>
-
     <!-- Test scope-->
     <dependency>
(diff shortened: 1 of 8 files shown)


pihme (approved):
Looks good to me. 

Please merge as soon as convenient. This will come in handy for me working on #227 and fighting against dependency checker.

PR #233, opened 2022-03-03
227 dependency checker

### LINKED ISSUE #227: Add maven dependency checker plugin

pihme:
## Description

- adds dependency checker
- fixes dependency errors

## Related issues

closes #227

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/extension/pom.xml
+++ b/extension/pom.xml
@@ -53,12 +53,12 @@
       <artifactId>slf4j-api</artifactId>
     </dependency>
 
-    <!-- Test scope-->
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-api</artifactId>
     </dependency>
 
+    <!-- Test scope-->
     <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
--- a/extension/pom.xml
+++ b/extension/pom.xml
@@ -71,6 +71,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <scope>test</scope>
+    </dependency>
+
   </dependencies>
   <build>
     <plugins>
--- a/extension/pom.xml
+++ b/extension/pom.xml
@@ -34,8 +34,18 @@
     </dependency>
 
     <dependency>
-      <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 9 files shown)


pihme:
@remcowesterhoud If you have some time, you could maybe take a look at this. Now the dependency checker is happy, but the tests are not running. There are two issues that I am aware of:

- the dependency checker removed all the log simple dependencies saying they were not needed. But I think you added them on purpose
- The testcontainers don't run now. I think this either has to do with the scope they were defined in, or maybe I need to add ignore statements to dependency checker.

I will take a second look after lunch. If you have any insights, please share.

pihme:
@remcowesterhoud Also why are the tests failing but the test results are all green?

remcowesterhoud:
> the dependency checker removed all the log simple dependencies saying they were not needed. But I think you added them on purpose

These are needed. They are not used directly in the code so that's probably why the dependency checker removes them. But we need to provide an slf4j implementation in our tests, otherwise we won't get any logging.

> The testcontainers don't run now. I think this either has to do with the scope they were defined in, or maybe I need to add ignore statements to dependency checker.

It seems the dependency checker removed the `java.annotation-api` dependency. Again this is a dependency that is not used directly in our code. However, it is used in the code that is generated by gRPC, which is why there are compile errors.

> Also why are the tests failing but the test results are all green?

Not quite sure what you mean, everything seems red to me?

pihme:
> Not quite sure what you mean, everything seems red to me?

https://github.com/camunda-cloud/zeebe-process-test/pull/233#issuecomment-1057928877

remcowesterhoud:
Hmm that's interesting. I think it's because the pipeline failed on compilation. Therefore no new artifacts with test reports are published. I guess that on publishing it will download the latest artifacts for this PR, which would be the ones of the previous commit 🤔

extension/pom.xml
```diff
@@ -45,16 +55,45 @@
 
     <!-- Test scope-->
     <dependency>
-      <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+    </dependency>
```
remcowesterhoud, line 60: 🔧 It mentions this dependency under test scope, but the scope of is not test. I don't think it should be test as this is in the extension, but I see there is also a new dependency `junit-platform-commons`. I'm not sure what this one  is.

remcowesterhoud (approved):
Looks good, just one small question/suggestion 👍

PR #234, opened 2022-03-03
Flaky test fixes

### LINKED ISSUE #202: Flaky test: ProcessEventInspectionsTest.testFindLastProcessInstance

remcowesterhoud:
### testLastProcessInstance

This test was flaky because we immediately increased the engine time after doing the deployment. This could result in the engine time being increased, before the timer is scheduled.
That means if the timer is expected to be scheduled in 1 day. It would now be scheduled in 1 day + the increase in engine time. This caused the timer to not be triggered at all.

### MultiThreadTest

I also encountered a different flaky test which is also fixed by this pr. This is the `MutltiThreadTest`.
This test was only flaky when running it using testcontainers. The problem lies in the `RecordStreamSourceWrapper`. In this class we would get all the records from the engine (`RecordStreamSource`). These records would be mapped to json and put into a list. The problem for this test was that we just override all records everytime we request them from the engine.

With this change we keep track of the latest event we've mapped. When we get a new request we will fetch all the records of the engine. We will filter those so we only have the records that we haven't stored in our list yet. These will be mapped and added. When the mapping is done we return a copy of the list. We need to make sure no more records are added, because the calling method will iterate over this list.

## Related issues

closes #202

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/engine-agent/pom.xml
+++ b/engine-agent/pom.xml
@@ -34,6 +34,15 @@
       <artifactId>slf4j-api</artifactId>
     </dependency>
 
+    <!--  Usually we don't want to provide an slf4j implementation in this project. This module
+      is the only exception to this rule. This is because we need to provide the engine with an
+      implementation, and since this runs in a docker container the user won't be providing it. -->
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-stub</artifactId>
@@ -79,6 +88,15 @@
             </configuration>
           </execution>
         </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <configuration>
+          <failOnWarning>true</failOnWarning>
+          <ignoreUnusedRuntime>true</ignoreUnusedRuntime>
+        </configuration>
       </plugin>
     </plugins>
   </build>
--- a/engine-agent/src/main/java/io/camunda/zeebe/process/test/engine/agent/RecordStreamSourceWrapper.java
+++ b/engine-agent/src/main/java/io/camunda/zeebe/process/test/engine/agent/RecordStreamSourceWrapper.java
@@ -2,12 +2,15 @@ package io.camunda.zeebe.process.test.engine.agent;
 
 import io.camunda.zeebe.process.test.api.RecordStreamSource;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.stream.StreamSupport;
 
 public class RecordStreamSourceWrapper {
 
   private final List<String> mappedRecords = new ArrayList<>();
   private final RecordStreamSource recordStreamSource;
+  private volatile long lastEventPosition = -1L;
 
   public RecordStreamSourceWrapper(final RecordStreamSource recordStreamSource) {
     this.recordStreamSource = recordStreamSource;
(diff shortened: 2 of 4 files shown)


engine-agent/pom.xml
```diff
@@ -34,6 +34,14 @@
       <artifactId>slf4j-api</artifactId>
     </dependency>
 
+    <!--  Usually we don't want to provide an slf4j implementation in this project. This modules
+      is the only exception to this rule. This is because we need to provide the engine with an
+      implementation, and since this runs in a docker container the user won't be providing it. -->
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
```
pihme, line 42: :wrench: I guess this could be runtime scope. And then you could configure dependency plugin to ignore runtime dependencies

engine-agent/src/main/java/io/camunda/zeebe/process/test/engine/agent/RecordStreamSourceWrapper.java
```diff
@@ -2,22 +2,31 @@
...
 
   public RecordStreamSourceWrapper(final RecordStreamSource recordStreamSource) {
     this.recordStreamSource = recordStreamSource;
   }
 
   public List<String> getMappedRecords() {
     synchronized (mappedRecords) {
-      mappedRecords.clear();
-      recordStreamSource.records().forEach(record -> mappedRecords.add(record.toJson()));
+      StreamSupport.stream(recordStreamSource.records().spliterator(), false)
```
pihme, line 21: :question: How does this work together with resetting the engine? Wouldn't resetting the engine also reset the record position to 0?

pihme (approved):
Looks good! Thanks for the persistence!

engine-agent/src/main/java/io/camunda/zeebe/process/test/engine/agent/RecordStreamSourceWrapper.java
remcowesterhoud: Yes it will. Resetting doesn't really reset the engine. It stops the old engine and recreate a new one. This also means there will be a fresh record stream, containing nothing.

pihme: If that is the case, shouldn't we also reset the RecordStreamWrapper when we reset the engine?

remcowesterhoud: A new instance of this class is created upon reseting, so there is no need to do this separately:
```java
@Override
  public void resetEngine(
      final ResetEngineRequest request,
      final StreamObserver<ResetEngineResponse> responseObserver) {
    engine.stop();
    engine = EngineFactory.create(AgentProperties.getGatewayPort());
    recordStreamSource = new RecordStreamSourceWrapper(engine.getRecordStreamSource());
    engine.start();

    final ResetEngineResponse response = ResetEngineResponse.newBuilder().build();
    responseObserver.onNext(response);
    responseObserver.onCompleted();
  }
```

PR #240, opened 2022-03-07
fix: enable google java format settings

remcowesterhoud:
## Description

Saving a file on IntelliJ would result in different formatting than applying spotless. This was because the google java format settings were not enabled.
This commit enables google java format, which should resolve the dicrepancy between IntelliJ and spotless.

## Definition of Done

Code changes:

Testing:

Documentation:


pihme (approved):
Oops

PR #248, opened 2022-03-09
chore: refactor worker test

### LINKED ISSUE #244: Engine.waitForIdleState() does not work as expected
## Description
Called for here: https://github.com/zeebe-io/zeebe-cluster-testbench/pull/591#discussion_r821394591

In that case, replacing `Thread.sleep(250)` with `engine.waitForIdleState()` causes the test to fail. Other instances have also been mentioned in the comment

### Expected behaviour

`waitForIdleState()` does what the name suggests

## Reproduction steps
- Check out code referred to in comment
- replace `Thread.sleep(250)` with `engine.waitForIdleState()` 

## Environment

- OS: Linux
- Version: 1.4.0-alpha2, 1.4.0-alpha3-SNAPSHOT (of March 8th)
- Flakiness observed in embedded and in testcontainer setup

pihme:
## Description
- remove Thread.sleep()
- wait until process instance gets completed

## Related issues

closes #244

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/qa/src/test/java/io/camunda/zeebe/process/test/qa/regular/multithread/WorkerTest.java
+++ b/qa/src/test/java/io/camunda/zeebe/process/test/qa/regular/multithread/WorkerTest.java
@@ -15,11 +15,13 @@
  */
 package io.camunda.zeebe.process.test.qa.regular.multithread;
 
+import static io.camunda.zeebe.process.test.assertions.BpmnAssert.assertThat;
+
 import io.camunda.zeebe.client.ZeebeClient;
-import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
+import io.camunda.zeebe.client.api.response.ProcessInstanceResult;
 import io.camunda.zeebe.process.test.api.ZeebeTestEngine;
-import io.camunda.zeebe.process.test.assertions.BpmnAssert;
 import io.camunda.zeebe.process.test.extension.ZeebeProcessTest;
+import io.camunda.zeebe.process.test.filters.RecordStream;
 import io.camunda.zeebe.process.test.qa.util.Utilities;
 import io.camunda.zeebe.process.test.qa.util.Utilities.ProcessPackLoopingServiceTask;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


pihme:
@remcowesterhoud Have a look and tell me what you think. I think it's definitely a step forward and hopefully removes flakiness.

This line is a bit awkward and not intuitive: `BpmnAssert.initRecordStream(recordStream);`

What could also help would be something like `engine.waitForJobCompletion(jobType);`

pihme:
We could also do a `startProcessInstanceWithResult` to make this example less non-sensical.

Btw, in the testbench test this will make sense read as a test, because it will be something like:
```
// when
startParentProcess();
await().until( parenProcessHasCompleted);
engine.waitForIdleState();

//then
assertThat(childProcess).isStarted().hasVariable(...)
```

remcowesterhoud:
I like using `withResult()` better than the await where you need to initialise the RecordStream. I don't want users to have to worry about this.

I liked the idea of adding a `waitForJobCompletion` but I think we should make it more generic. We could make a `waitFor(ValueType, Intent)`. That way we can wait for everything with a single method.

pihme:
> I like using `withResult()` better than the await where you need to initialise the RecordStream. I don't want users to have to worry about this. Though, I'm not sure what happens when the process doesn't complete but runs in an incident.

This will produce a request timeout exception.

> I liked the idea of adding a `waitForJobCompletion` but I think we should make it more generic. We could make a `waitFor(ValueType, Intent)`. That way we can wait for everything with a single method.

Maybe a little too technical, but a good idea.

pihme:
I now have implemented only the refactoring for the test. The additional methods on `Engine` could be a follow up issue, independently prioritized.

remcowesterhoud (changes requested):
Thanks! I like the changes so much that I'd like to see them a 2nd time in the testcontainer version of this test 😉 (this duplication is already annoying me so this will be one of the first things I want to get rid of next).

I've created a separate issue for the `waitFor` method: https://github.com/camunda-cloud/zeebe-process-test/issues/252

remcowesterhoud (approved):
LGTM!

PR #255, opened 2022-03-11
Remove test duplication

### LINKED ISSUE #195: Remove QA test duplication
**Description**

Currently we have our QA tests duplicated. This is because we want to run them once using the embedded engine, and once using testcontainers. We should look for a way to reuse the same tests, whilst still running it using the different extensions.

Possible interesting sources:
- https://junit.org/junit5/docs/current/user-guide/#extensions-registration-automatic
- https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html

remcowesterhoud:
## Description

This PR introduces abstract test classes. Abstract test classes should contain all the test implementations. The other classes that extend this only need to provide the extension. This way we don't have to duplicate all of our test implementations.

One thing to note is that injection of the `ZeebeClient`, `ZeebeTestEngine` and `RecordStream`. These fields get injected in what jUnit defines as the test class:
`final Class<?> requiredTestClass = extensionContext.getRequiredTestClass();`

We are dealing with 2 different scenarios:
1. We have the assertion tests. Here the abstract classes contain 2 nested classes, `HappyPathTests` and `UnhappyPathTests`. In this scenario the nested class is the test class. These fields get injected just fine.
2. In other tests we do not have any nested classes. Here the class that extends the abstract class will be the test class (e.g. the `WorkterTest` class). For these classes the fields must be declared on the extending class. We can access these in the abstract class by adding abstract getters.

## Related issues

closes #195 

## Definition of Done

Code changes:

Testing:

Documentation:


qa/src/test/java/io/camunda/zeebe/process/test/qa/ExtendAbstractsTest.java
```diff
@@ -0,0 +1,94 @@
...
+  private static final String ABSTRACTS = "abstracts";
+
+  @Test
+  void testAbstractClassesAreExtendedWithBothExtensions() throws IOException {
+    final Map<String, List<Class<?>>> classes = new HashMap<>();
+    classes.put(ABSTRACTS, new ArrayList<>());
+    classes.put(EMBEDDED, new ArrayList<>());
+    classes.put(TESTCONTAINER, new ArrayList<>());
+
+    ClassPath.from(ClassLoader.getSystemClassLoader()).getTopLevelClasses().stream()
```
pihme, line 48: :wrench: This seems correct, but also hard to read. I would be strongly in favor of using something like https://github.com/ronmamo/reflections or https://github.com/classgraph/classgraph. It's maybe overkill for what you are trying to achieve. But my hope is that it would be easier to comprehend. And you could also make the tests more precise, e.g. abstract classes should be extended, and they should have a certain annotation on them, or something like that.

pihme (approved):
Looks clean. One suggestion to make it even cleaner

remcowesterhoud:
@pihme I've changed the test to use `ClassGraph`. Please have another look and see if it's improved. I quite like it 😄

qa/src/test/java/io/camunda/zeebe/process/test/qa/ExtendAbstractsTest.java
```diff
@@ -15,80 +15,76 @@
...
-                classes.get(ABSTRACTS).add(classInfo.load());
-              } else if (classInfo.getPackageName().contains(BASE_PACKAGE + "." + EMBEDDED)) {
-                classes.get(EMBEDDED).add(classInfo.load());
-              } else if (classInfo.getPackageName().contains(BASE_PACKAGE + "." + TESTCONTAINER)) {
-                classes.get(TESTCONTAINER).add(classInfo.load());
-              }
-            });
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("provideAbstractClasses")
+  void testAbstractClassIsExtendedWithBothExtensions(
```
pihme, line 50: :wrench: could be split into two tests

qa/src/test/java/io/camunda/zeebe/process/test/qa/ExtendAbstractsTest.java
```diff
@@ -15,80 +15,76 @@
...
+      final String className, final Class<?> abstractClass) {
+    final ClassInfoList embeddedClass =
+        EXTENDING_CLASSES
+            .filter(info -> info.getPackageName().contains("embedded"))
```
pihme, line 54: :wrench: could use `EMBEDDED_PACKAGE` dito the other one

qa/src/test/java/io/camunda/zeebe/process/test/qa/ExtendAbstractsTest.java
```diff
@@ -15,80 +15,76 @@
...
+  private static final String TESTCONTAINER_PACKAGE =
+      "io.camunda.zeebe.process.test.qa.testcontainer";
 
-  @Test
-  void testAbstractClassesAreExtendedWithBothExtensions() throws IOException {
-    final Map<String, List<Class<?>>> classes = new HashMap<>();
-    classes.put(ABSTRACTS, new ArrayList<>());
-    classes.put(EMBEDDED, new ArrayList<>());
-    classes.put(TESTCONTAINER, new ArrayList<>());
+  private static final ClassInfoList EXTENDING_CLASSES =
```
pihme, line 39: :wrench: Not really needed. You put two packages into one bucket and then later filter out by package. You could as well prepare two buckets.

pihme (approved):
:tada: Looks much better already. Some final suggestions for the cherry on top

PR #264, opened 2022-03-24
Bump Zeebe to version 8.0.0-rc1

### LINKED ISSUE #262: Adjust ColumnFamily implementation to reflect API changes in Zeebe
**Description**

In Zeebe the API for ColumnFamily has changed. Instead of `put` there is now `update`, `insert`, `upsert`. Instead of `delete` there is now `deleteExisting` and `deleteIfExists`  

This change needs to be reflected here.

In addition to that, we need to decide whether we want to replicate the integrity checks. 

@pihme IMO we should wait with implementing the integrity checks until they are enabled by default in Zeebe.

remcowesterhoud:
## Description

The ColumnFamily interface in Zeebe has been modified. Before it only contained a put method. Starting from version 8.0.0-rc1 this has been replaced with update, insert and upsert methods.
The delete method has been replaced with deleteExisting and deleteIfExists.

In our in memory database we implement this interface. When updating to Zeebe 8.0.0-rc1 these changed need to be merged to keep the code from breaking.

In addition to the ColumnFamily changes the zeebe-protocol-jackson changed in the way we need to do the mapping. This minor mapping change has also been fixed in this PR.

## Related issues

closes #262

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/db/InMemoryDbColumnFamily.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/db/InMemoryDbColumnFamily.java
@@ -7,7 +7,12 @@
  */
 package io.camunda.zeebe.process.test.engine.db;
 
-import io.camunda.zeebe.db.*;
+import io.camunda.zeebe.db.ColumnFamily;
+import io.camunda.zeebe.db.DbKey;
+import io.camunda.zeebe.db.DbValue;
+import io.camunda.zeebe.db.KeyValuePairVisitor;
+import io.camunda.zeebe.db.TransactionContext;
+import io.camunda.zeebe.db.ZeebeDbInconsistentException;
 import io.camunda.zeebe.util.buffer.BufferUtil;
 import java.util.Iterator;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


pihme (changes requested):
This is a hard one. The changes look ok. But now the behavior of InMemoryEngine differs from Zeebe. In Zeebe the consistency checks are governed by feature flags. Currently the checks are disabled by default.

The tricky bit is that we don't have any notion of feature flags yet in the InMemoryEngine. And maybe we should not? I am torn on this one, to be honest.

I also wonder whether the foreign key checks are in place because they happen on another level, or whether we have to reimplement those as well.

I see two ways going forward:
* we allow to pass in a configuration object and reimplement the relevant feature flags here
* we purposefully deviate from the behavior of Zeebe, but then I would at least try to mimick the default configuration, which currently is with integrity checks disabled.

There should also be tests in Zeebe for behavior of this class with feature flags enabled vs disabled. Maybe we should copy those as well.

remcowesterhoud:
After some discussion we have decided to keep the changes as they are at this moment. This means we will check for existing keys where relevant, and we will not check for foreign keys.

The reasoning behind this decision is that the `ForeignKeyChecker` in Zeebe is not reusable in this project currently. If we would add this consistency check it would involve creating it ourselves. This is doable. However, with creating our own implementation we are already deviating from Zeebe. Doing this will raise the question what the value of this check is. It will differ from the Zeebe implementation and it is not something users of this project will require us to have. Any conflict on our foreign key check will be short-lived as the db will be reset after each test case anyways.

For this reason we have decided to deviate from the Zeebe database. We aim to keep the databases as similar as possible, but this is not always a feasible option.

PR #284, opened 2022-03-31
Explicitly depend on google-java-format

### LINKED ISSUE #277: google-java-format is not kept up-to-date
## Description

See https://github.com/camunda/zeebe/issues/9024

### Expected behaviour

Dependabot creates a PR to update the google-java-format version.

korthout:
## Description

Dependabot does not control the version of the google-java-format plugin used by spotless. By adding an explicit maven dependency on the plugin we can make Dependabot control it.

Note that adding this dependency doesn't change anything to the rest of the project. It also doesn't really interfere with this specific plugin, as it would in the worst case, simply add the same classes twice on the classpath. By using a property, we can control that the dependent version is the same as the one loaded inside spotless.

## Related issues

closes #277 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/.idea/externalDependencies.xml
+++ b/.idea/externalDependencies.xml
@@ -2,6 +2,6 @@
 <project version="4">
   <component name="ExternalDependencies">
     <plugin id="com.dubreuia" />
-    <plugin id="google-java-format" />
+    <plugin id="google-java-format" min-version="1.12.0.0"/>
   </component>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -421,6 +421,13 @@
               </sortPom>
             </pom>
           </configuration>
+          <dependencies>
+            <dependency>
+              <groupId>com.google.googlejavaformat</groupId>
+              <artifactId>google-java-format</artifactId>
+              <version>${plugin.version.googlejavaformat}</version>
+            </dependency>
+          </dependencies>
           <executions>
             <execution>
               <goals>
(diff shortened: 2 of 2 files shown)


.idea/externalDependencies.xml
```diff
@@ -2,6 +2,6 @@
 <project version="4">
   <component name="ExternalDependencies">
     <plugin id="com.dubreuia" />
-    <plugin id="google-java-format" />
+    <plugin id="google-java-format" min-version="1.12.0.0"/>
```
remcowesterhoud, line 5: 🔧 Please add a comment specifying the extra .0 is required to prevent confusion

remcowesterhoud (approved):
First contribution 🎉 Thanks @korthout!

I've added a similar comment as I did in the Zeebe project. Please have a look at it and merge it after. Small tip: we don't have bors here 😄

korthout:
>Small tip: we don't have bors here 😄

❓ Why do we not have bors here?

.idea/externalDependencies.xml
korthout: I chose not to add the comment, see https://github.com/camunda/zeebe/pull/9032#discussion_r840553278

remcowesterhoud:
Because I haven't set it up. I see it's value I just haven't felt the need for it yet since I was basically working solo on this project for the majority of the time. That way it is easy to prevent semantic conflicts.

PR #293, opened 2022-04-05
chore(pom): increase zeebe to 8.0.0

pihme:
## Description

Increases Zeebe version to 8.0.0


pihme:
@remcowesterhoud I will skip the release process this time. Since this is the first release, I feel we don't have the routine right just now. I read the description of steps to do in Zeebe Release process. Still, I am unsure when to update the version of Zeebe process test (i.e. before or after the tag). So please, review it, and then proceed with the release. I will continue with the main Zeebe release in the meantime.

PR #288, opened 2022-04-01
Run testcontainer tests with java 8

remcowesterhoud:
## Description

We claim to have the guarantee that the `zeebe-process-test-extension-testcontainer` is java 8 compatible. We are guaranteeing this by setting the java version in the maven configuration. However, we do not test this in our CI. The testcontainer tests we have are still being run with java 17.

This PR changes this. By using maven profiles to manage dependencies and plugins we can compile and run the testcontainer tests using java 8. With this we can test if our guarantee holds any truth.

## Related issues

closes #287

## Definition of Done

_Not all items need to be done depending on the issue and the pull request._

Code changes:
* [x] The changes are backwards compatibility with previous versions
* [ ] If it fixes a bug then PRs are created to backport the fix

Testing:
* [x] There are unit/integration tests that verify all acceptance criterias of the issue
* [ ] New tests are written to ensure backwards compatibility with further versions
* [ ] The behavior is tested manually

Documentation:
* [ ] Javadoc has been written
* [ ] The documentation is updated

PR #306, opened 2022-04-11
Run tests using jdk8

### LINKED ISSUE #287: Ensure running tests against actual Java 8 in workflow

remcowesterhoud:
## Description

We claim to have the guarantee that the `zeebe-process-test-extension-testcontainer` is java 8 compatible. We are guaranteeing this by setting the java version in the maven configuration. However, we do not test this in our CI. The testcontainer tests we have are still being run with java 17.

This has been done by splitting the QA module into 3 separate modules. 1 for the abstract tests, 1 for the embedded tests and 1 for the testcontainer tests. By excluding / including specific modules we can configure what tests we want to run in which scenario.

Please note: this PR has been based off of #288. Here we did the same thing using a different approach. Instead of restarting from the beginning I based this solution off of this branch and modified what was necessary. This is why the first commits are introducing the maven profiles. The relevant commits in this PR are the commits from the 11th of April.
 To me using these modules are a more maintainable approach compared to the profiles.

## Related issues

closes #287

## Definition of Done

Code changes:

Testing:

Documentation:


remcowesterhoud:
@saig0 I have changed the solution of #288 to use modules instead of profiles after discussing with Nicolas. I believe this approach is more clear and better to maintain. Since you had a look at the previous, would you mind taking a look at this one as well?

saig0 (approved):
@remcowesterhoud looks good :+1:

PR #307, opened 2022-04-11
NPE during logging

### LINKED ISSUE #305: NPE in RecordStreamLogger.logVariables(
Running my tests in https://github.com/camunda-community-hub/camunda-cloud-examples/blob/c8/twitter-review-java-springboot/src/test/java/org/camunda/community/examples/twitter/TestTwitterProcess.java I just saw this NPE (which should simply not happen I guess) on `8.0.0`

```log
java.lang.NullPointerException: Cannot invoke "Object.toString()" because "value" is null
	at io.camunda.zeebe.process.test.filters.logger.RecordStreamLogger.lambda$logVariables$9(RecordStreamLogger.java:275)
	at java.base/java.util.HashMap.forEach(HashMap.java:1421)
	at io.camunda.zeebe.process.test.filters.logger.RecordStreamLogger.logVariables(RecordStreamLogger.java:275)
	at io.camunda.zeebe.process.test.filters.logger.RecordStreamLogger.logProcessInstanceCreationRecordValue(RecordStreamLogger.java:238)
	at io.camunda.zeebe.process.test.filters.logger.RecordStreamLogger.lambda$logRecords$7(RecordStreamLogger.java:104)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1092)
	at io.camunda.zeebe.process.test.filters.logger.RecordStreamLogger.logRecords(RecordStreamLogger.java:100)
	at io.camunda.zeebe.process.test.filters.logger.RecordStreamLogger.log(RecordStreamLogger.java:93)
	at io.camunda.zeebe.process.test.filters.RecordStream.print(RecordStream.java:125)
	at io.camunda.zeebe.spring.test.ZeebeTestExecutionListener.afterTestMethod(ZeebeTestExecutionListener.java:39)
```

remcowesterhoud:
## Description

Variable with a null value produce NPE when logging the record stream. This PR adds a null check and a bunch of tests for our logger to prevent this in the future.

## Related issues

closes #305 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
+++ b/filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
@@ -266,20 +266,13 @@ public class RecordStreamLogger {
     return joiner.toString();
   }
 
-  private String logVariables(final Map<String, Object> variables) {
+  protected String logVariables(final Map<String, Object> variables) {
     if (variables.isEmpty()) {
       return "";
     }
 
     final StringJoiner joiner = new StringJoiner(", ", "[", "]");
-    variables.forEach(
-        (key, value) -> {
-          if (value != null) {
-            joiner.add(key + " -> " + value);
-          } else {
-            joiner.add(key + " -> null");
-          }
-        });
+    variables.forEach((key, value) -> joiner.add(key + " -> " + value));
     return String.format("(Variables: %s)", joiner);
   }
 
--- a/filters/src/test/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLoggerTest.java
+++ b/filters/src/test/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLoggerTest.java
@@ -15,15 +15,33 @@
  */
 package io.camunda.zeebe.process.test.filters.logger;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 import io.camunda.zeebe.protocol.record.Record;
 import io.camunda.zeebe.protocol.record.ValueType;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Function;
+import java.util.stream.Stream;
 import org.assertj.core.api.SoftAssertions;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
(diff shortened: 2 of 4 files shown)


filters/src/test/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLoggerTest.java
```diff
@@ -15,16 +15,56 @@
...
+import java.util.Date;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Function;
 import org.assertj.core.api.SoftAssertions;
 import org.junit.jupiter.api.Test;
 
 class RecordStreamLoggerTest {
 
+  private static final Map<String, Object> TYPED_TEST_VARIABLES = new HashMap<>();
```
pihme, line 58: :x: I am not so sure these tests add much value. I think it would be better to make `logVariables(...)` in `RecordStreamLogger` protected and test this method in isolation. 

if you still want to keep those kinds of tests I would propose to go deep into jqwik and make "Does not throw an exception" a property and then refactor this to spend most of the time generating aribtrary records. I think this will be a better investment going forward

filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
```diff
@@ -272,7 +272,14 @@ private String logVariables(final Map<String, Object> variables) {
     }
 
     final StringJoiner joiner = new StringJoiner(", ", "[", "]");
-    variables.forEach((key, value) -> joiner.add(key + " -> " + value.toString()));
+    variables.forEach(
```
pihme, line 275: :wrench: This could also be rewritten to (pseudo code)
```
variables.entries().stream()
   .map( (key, value) -> key + " -> " +  value == null ? "null" :  value )
   .collect(Collectors.joining(", "));
```

pihme (changes requested):
Change looks good. I think the tests should be looked at again.

filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
pihme: Or even simpler:
` variables.forEach((key, value) -> joiner.add(key + " -> " + value()));`

pihme (approved):

PR #319, opened 2022-04-13
Support sequence flow in hasPassedElement assertions

### LINKED ISSUE #318: Support sequence flows in hasPassedElement and hasPassedElementsInOrder
## Description

It could be valuable for a user to test if they have taken a specific sequence flow. Currently we have no way to do this. Users would expect `hasPassedElement` and `hasPassedElementsInOrder` could be used for this, however these assertions do not include sequence flow.

### Expected behaviour

`hasPassedElement` and `hasPassedElementsInOrder` should be able to assert sequence flows.

### Solution
Both methods have a similar filter:

```java
 StreamFilter.processInstance(recordStream)
            .withProcessInstanceKey(actual)
            .withRejectionType(RejectionType.NULL_VAL)
            .withElementIdIn(elementIds)
            .withIntent(ProcessInstanceIntent.ELEMENT_COMPLETED)
```

The intent here causes sequence flows to be excluded, as we never write an `ELEMENT_COMPLETED` event for these. We should add a way to filter for multiple intents and turn this into:
```java
.withIntents(ProcessInstanceIntent.ELEMENT_COMPLETED, ProcessInstanceIntent.SEQUENCE_FLOW_TAKEN)
```

remcowesterhoud:
## Description

Sequence flows will never have an event with an ELEMENT_COMPLETED intent. Instead they get a single event with a SEQUENCE_FLOW_TAKEN intent. Because of this we would always filter out any sequence flows in the assertion, making it impossible to assert that a specific flow has been passed. By adding the SEQUENCE_FLOW_TAKEN intent we include sequence flows in our lists and allow asserting whether or not they have been passed.

## Related issues

closes #318 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/assertions/src/main/java/io/camunda/zeebe/process/test/assertions/ProcessInstanceAssert.java
+++ b/assertions/src/main/java/io/camunda/zeebe/process/test/assertions/ProcessInstanceAssert.java
@@ -214,7 +214,8 @@ public class ProcessInstanceAssert extends AbstractAssert<ProcessInstanceAssert,
             .withProcessInstanceKey(actual)
             .withRejectionType(RejectionType.NULL_VAL)
             .withElementId(elementId)
-            .withIntent(ProcessInstanceIntent.ELEMENT_COMPLETED)
+            .withIntents(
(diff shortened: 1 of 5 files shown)


saig0 (approved):
@remcowesterhoud LGTM :+1:

PR #348, opened 2022-05-03
build: set zeebe version to 8.1.0-alpha1

deepthidevaki:
## Description

Set zeebe version to 8.1.0-alpha1 as part of the zeebe release process

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -223,6 +223,13 @@
         <version>${dependency.zeebe.version}</version>
       </dependency>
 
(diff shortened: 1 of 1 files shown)


deepthidevaki:
@remcowesterhoud I have opened an issue for this in zeebe https://github.com/camunda/zeebe/issues/9280

To temporarily fix this, shall we pin the version to a specific one? Or is there a way to ignore this dependency in dependency analysis.

remcowesterhoud (approved):
👍 LGTM once the tests succeeded

PR #389, opened 2022-06-02
Run testcontainers based tests against TC Cloud

npepinpe:
## Related issues

closes #

## Definition of Done

Code changes:

Testing:

Documentation:


npepinpe:
Seems to be slower with TCC. Uh, I wasn't expecting that. At worst I was expecting the same, since there's only 2 vCPUs so you can't parallelize much, but it seems just generally slower. Ah well, let's just drop it then.

PR #399, opened 2022-06-14
Dedicated example

### LINKED ISSUE #367: Examples are hard to understand
**Description**

All examples just extend an abstract class with the test implementation. 

As a new user looking for examples it's quite difficult to understand, especially if you look up the code on github directly in your browser.

Easy examples without full featured DRY would be much better to learn from.

pihme:
## Background

Initially we wanted to use our own tests also as examples to avoid duplicating code. However, our own tests are not ideal as examples:
* They are implemented once and then run against different test engine implementations in different versions, which is hard to understand, and also not necessary to understand by users
* They include logic that is based on interpreting the log records returned by the engine. This is common in our tests, but not how we recommend users to write their tests
* They are exhaustive, but often only include a single assertion

We also have a more complex tests, which uses different assertions in combination. But this test was hard to find among all the other files and also suffered from aforementioned problems.

## Change
This PR
* introduces a new module examples and advertises it in the readme
* Moved the complex test into that new module so that it stands out more
* Inlined code so that the test is only a single file
* Added comments to explain details in the test which are less obvious
* Rewrote those parts of the code that used exported records to only use Zeebe Client API commands

## Related issues

closes #367

## Definition of Done

Code changes:

Testing:

Documentation:
* [X] The documentation is updated


saig0 (approved):
@pihme I like it :+1: Good job :tada:

PR #408, opened 2022-06-16
Build docker image with Maven

### LINKED ISSUE #407: Build Docker image as part of Maven build
**Description**

Currently the QA tests for testcontainer extension depend on an existing Docker image. This Docker image is currently built outside of Maven. 

**Solution**
It would be good, if the Maven build would also build the Docker image that is needed for the QA tests.

pihme:
## Description

For local builds (on a developer's workstation) a Docker image is built by Maven on the fly. This image is then used in QA tests with testcontainers. All tests locally will be run with Java 17.

The GitHub workflows should be unperturbed by this change.

## Review hints
Please have a look at the issue for the intended scope of the change.

I think there could be further steps to also use Maven to build the Docker image as part of the GHA workflows, but I wanted to keep the change set small.

Now, effectively Maven will build a local Docker image that is tagged with the project version and GHA for build and test will build a Docker image that is tagged with the SHA value. So the two shouldn't conflict with each other :crossed_fingers: 

I was also wondering whether we should document this https://github.com/camunda/zeebe-process-test/issues/407#issuecomment-1157568109 as part of some developer documentation. Basically, I always need an hour or two to wrap my head around how this works in detail. It is easy to see from the GHA workflows what is happening, but not so easy to piece together why it needs to happen in that way. Please let me know what you think.

This will also use the default time stamping for the created images. Please read the reasoning here: https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#why-is-my-image-created-48-years-ago I opted not to deviate from the default. Again, please let me know what you think

## Related issues

closes #407 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -49,7 +49,7 @@ jobs:
 
       - name: Build
         id: build
-        run: mvn -B -U -pl "-:zeebe-process-test-qa-testcontainers" "-Dsurefire.rerunFailingTestsCount=5" clean install
+        run: mvn -B -U -pl "-:zeebe-process-test-qa-testcontainers" -P \!localBuild "-Dsurefire.rerunFailingTestsCount=5" clean install
 
       - name: Archive Test Results
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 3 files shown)


remcowesterhoud (approved):
🚀 Great change

PR #464, opened 2022-08-03
Daily test against Zeebe SNAPSHOT version

### LINKED ISSUE #465: Test daily against Zeebe SNAPSHOT version
**Description**

We've encountered problems during the release multiple times, because Zeebe contained breaking changed for Zeebe Process Test. By building and running our tests we can identify these problems earlier on in the development process and don't have to fix these during the release process.

If the build failed we should send a notification to the `#zeebe-ci` slack channel.

remcowesterhoud:
## Description

We've encountered problems during the release multiple times, because Zeebe contained breaking changes for Zeebe Process Test. By building and running our tests we can identify these problems earlier on in the development process and don't have to fix these during the release process.

Logically we have to decide in the workflow what the next Zeebe SNAPSHOT version will be from the current Zeebe version defined in the pom. There is 2 options for the current version:

1. No alpha version: MAJOR.MINOR.PATCH
2. Alpha version: MAJOR.MINOR.PATCH-alphaX

When it is not an alpha version we need to increment the MINOR version by 1 and append `-SNAPSHOT` at the end. When it is an alpha version we need to replace `alphaX` with `SNAPSHOT`.

## Related issues

closes #465

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/.github/workflows/snapshot-test.yml
+++ b/.github/workflows/snapshot-test.yml
@@ -38,3 +38,38 @@ jobs:
           name: Unit Test Results Against Zeebe Snapshot
           path: "*/target/surefire-reports/"
           retention-days: 7
+
+  notify-if-failed:
+    name: Send slack notification on build failure
+    runs-on: ubuntu-latest
+    needs: [test-against-zeebe-snapshot]
+    if: failure()
+    steps:
+      - id: slack-notify
+        name: Send slack notification
+        uses: slackapi/slack-github-action@v1.21.0
+        with:
+          # For posting a rich message using Block Kit
+          payload: |
+            {
+              "text": ":alarm: `zeebe-process-test` build against Zeebe SNAPSHOT version failed! :alarm:",
+             	"blocks": [
+                {
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": ":alarm: `zeebe-process-test` build against Zeebe SNAPSHOT version failed! :alarm:"
+                  }
+                },
+                {
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": "Please check the related workflow (${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}) and fix it before it blocks the upcoming release!\n  \\cc @zeebe-medic"
+                  }
+                }
+              ]
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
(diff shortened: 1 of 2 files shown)


.github/workflows/snapshot-test.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Build with Zeebe SNAPSHOT version
+        run: |
+          CURRENT_ZEEBE_VERSION=$(mvn help:evaluate -Dexpression=dependency.zeebe.version -DforceStdout -q)
+          if grep -q "alpha" <<< "$CURRENT_ZEEBE_VERSION"; then
+            NEW_ZEEBE_VERSION=$(echo $CURRENT_ZEEBE_VERSION | sed "s/alpha.*/SNAPSHOT/")
+          else
+            NEW_ZEEBE_VERSION=$(echo $CURRENT_ZEEBE_VERSION | awk -F. '/[0-9]+\./{$2++;print}' OFS=.)
+          fi
+          echo $NEW_ZEEBE_VERSION
+          mvn -B -U -pl "-:zeebe-process-test-qa-testcontainers" -P !localBuild "-Dsurefire.rerunFailingTestsCount=5" -Ddependency.zeebe.version=$NEW_ZEEBE_VERSION clean install
```
saig0, line 32: Too much command line magic for me :dizzy_face:

saig0 (approved):
@remcowesterhoud nice :+1: I like the idea :rocket: 

image

.github/workflows/snapshot-test.yml
remcowesterhoud: It was a lot of googling for me too 😅 

Long story short what it does:
1. Read the `dependency.zeebe.version` from the pom file and store it in the `CURRENT_ZEEBE_VERSION` variable.
2. If the `CURRENT_ZEEBE_VERSION` is an alpha version it creates the `NEW_ZEEBE_VERSION` variable. This gets the value of the `CURRENT_ZEEBE_VERSION`, but replaces alpha and everything that comes after this with `SNAPSHOT`. E.g. `8.1.0-alpha4` becomes `8.1.0-SNAPSHOT`.
3. If it's not an alpha version we do some actual magic I stole from a StackOverflow post 🪄 It takes the `CURRENT_ZEEBE_VERSION` and from what I understand it groups the version number into 3 sections each divided by a dot. So that would be the major, minor and patch version. `{$2++;print}` tells us to get the 2nd group (the minor version) and increment it with 1. I have no clue what `OFS=.` does.
4. Afterwards we just run the tests using maven overriding the zeebe version with `NEW_ZEEBE_VERSION` (`-Ddependency.zeebe.version=$NEW_ZEEBE_VERSION`)

saig0: Thank you for explaining :tophat: :magic_wand:

PR #471, opened 2022-08-08
Apply recent engine refactorings

Zelldon:
## Description

Adjust the code base to the recent engine abstraction refactorings

## Related issues

closes #

## Definition of Done

_Not all items need to be done depending on the issue and the pull request._

Code changes:
* [ ] The changes are backwards compatibility with previous versions
* [ ] If it fixes a bug then PRs are created to backport the fix

Testing:
* [ ] There are unit/integration tests that verify all acceptance criterias of the issue
* [ ] New tests are written to ensure backwards compatibility with further versions
* [ ] The behavior is tested manually

Documentation:
* [ ] Javadoc has been written
* [ ] The documentation is updated

PR #489, opened 2022-09-07
build(deps): update zeebe version to 8.1.0-alpha5

deepthidevaki:
## Description

Update zeebe version to 8.1.0-alpha5

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,7 +73,7 @@
     <dependency.snakeyaml.version>1.31</dependency.snakeyaml.version>
     <dependency.spring.version>5.3.22</dependency.spring.version>
     <dependency.testcontainers.version>1.17.3</dependency.testcontainers.version>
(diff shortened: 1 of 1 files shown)


korthout (changes requested):
Thanks @deepthidevaki 

The last time this was changed to `SNAPSHOT` to accommodate the changes from the engine abstraction topic, see:
- #471

I think it makes sense that we update the version for a new release, but that shouldn't happen on `main`. Instead, we should probably cut a release by branching and then change the version on that branch. We can then release a version from that branch.

If that doesn't work with the current CI pipelines, then we can make the change to `main` as you've proposed here, **and** later rollback to point to SNAPSHOT again for continued development.

❌ Please adjust the target branch.

korthout (approved):
It seems I was making a wrong assumption. The release does happen on `main`.

In that case, this looks good to me.

🙇 Please make sure that this change is reverted after the release, to support continued development.

PR #500, opened 2022-09-15
Fully support Process Instance Modification

### LINKED ISSUE #466: Support the process instance modification command
## Description

With Zeebe `8.1.0`, we added a new process instance modification API. We should support the new command also in ZPT.

- Map the request to a command in the `GrpcToLogStreamGateway` and write it
- Map the response in the `GrpcResponseMapper`

remcowesterhoud:
## Description

This PR adds support for the PI modification command by adding it to the ZPT gateway. Handling the incoming request was already taken care of in #461.
With this PR we now map the Response of the request as well. For modification this is an empty response. It also adds a testcase to verify that modifying a process instance works. This test is not meant as an exhaustive test for the modification logic. What we're interested in is making sure the request is handled properly by the gateway.

## Related issues

closes #466 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
+++ b/engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
@@ -42,6 +42,7 @@ import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 import org.awaitility.Awaitility;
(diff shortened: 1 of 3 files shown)


engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
```diff
@@ -332,6 +333,78 @@ void shouldCancelProcessInstance() {
...
+
+    zeebeEngine.waitForIdleState(Duration.ofSeconds(1));
+
+    final long task1Key =
+        StreamSupport.stream(
+                RecordStream.of(zeebeEngine.getRecordStreamSource())
+                    .processInstanceRecords()
+                    .spliterator(),
+                false)
+            .filter(r -> r.getIntent() == ProcessInstanceIntent.ACTIVATE_ELEMENT)
```
korthout, line 368: ❓ Could it be that this was a typo and instead of `ACTIVATE_ELEMENT` it was intended to find the first `ELEMENT_ACTIVATED`?

korthout (approved):
Nice work @remcowesterhoud

LGTM 👍

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
remcowesterhoud: You are correct. It shouldn't matter functionality-wise, but I'll change it either way. Good catch!

PR #526, opened 2022-10-05
Fix version decision command

remcowesterhoud:
## Description

The command to increase the Zeebe minor by 1 and appending -SNAPSHOT was broken. `            NEW_ZEEBE_VERSION=echo $(echo $CURRENT_ZEEBE_VERSION | awk -F. '/[0-9]+\./{$2++;print}' OFS=.)-SNAPSHOT` would result in bash trying to execute the version. This would result in the message that 8.2.0-SNAPSHOT is not a command:
image

By moving the $() outward and using backticks this problem is resolved. `NEW_ZEEBE_VERSION=$(echo `echo $CURRENT_ZEEBE_VERSION | awk -F. '/[0-9]+\./{$2++;print}' OFS=.`-SNAPSHOT)` will echo the version instead of trying to execute it.
image

## Related issues

closes #

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/.github/workflows/snapshot-test.yml
+++ b/.github/workflows/snapshot-test.yml
@@ -28,7 +28,7 @@ jobs:
           elif grep -q "SNAPSHOT" <<< "$CURRENT_ZEEBE_VERSION"; then
             NEW_ZEEBE_VERSION=$CURRENT_ZEEBE_VERSION
           else
-            NEW_ZEEBE_VERSION=echo $(echo $CURRENT_ZEEBE_VERSION | awk -F. '/[0-9]+\./{$2++;print}' OFS=.)-SNAPSHOT
+            NEW_ZEEBE_VERSION=`echo $CURRENT_ZEEBE_VERSION | awk -F. '/[0-9]+\./{$2++;print}' OFS=.`-SNAPSHOT
           fi
           echo $NEW_ZEEBE_VERSION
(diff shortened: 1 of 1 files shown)


.github/workflows/snapshot-test.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           elif grep -q "SNAPSHOT" <<< "$CURRENT_ZEEBE_VERSION"; then
             NEW_ZEEBE_VERSION=$CURRENT_ZEEBE_VERSION
           else
-            NEW_ZEEBE_VERSION=echo $(echo $CURRENT_ZEEBE_VERSION | awk -F. '/[0-9]+\./{$2++;print}' OFS=.)-SNAPSHOT
+            NEW_ZEEBE_VERSION=$(echo `echo $CURRENT_ZEEBE_VERSION | awk -F. '/[0-9]+\./{$2++;print}' OFS=.`-SNAPSHOT)
```
oleschoenburg, line 31:
```suggestion
            NEW_ZEEBE_VERSION=`echo $CURRENT_ZEEBE_VERSION | awk -F. '/[0-9]+\./{$2++;print}' OFS=.`-SNAPSHOT
```

seems to work too

remcowesterhoud: Thanks I like this better 👍

PR #584, opened 2022-11-30
fix: adjust logstream writer usage after refactoring

oleschoenburg:
## Description

Following the refactoring in https://github.com/camunda/zeebe/pull/11115, I've updated the usage of the logstream writer API


Zelldon (approved):
Thanks 🚀👍

PR #589, opened 2022-12-06
fix(ci): Switch back to actions/setup-java@v3.6.0

saig0:
## Description

The version `v3.7.0` of actions/setup-java was removed. Fix the workflows by using the previous version. See https://github.com/actions/setup-java/issues/422.

## Related issues

Failed action: https://github.com/camunda/zeebe-process-test/actions/runs/3626043207

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set up Java environment
-        uses: actions/setup-java@v3.7.0
+        uses: actions/setup-java@v3.6.0
         with:
           distribution: temurin
           java-version: 17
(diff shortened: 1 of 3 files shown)


remcowesterhoud (approved):
Interesting decision by them 😅

bors merge

korthout:
It might be better to switch the version to the latest major release `v3`. Such tags are generally moved with every new minor and patch release of the major version. This practice is officially recommended by GH.

Note that this may result in reduced reproducibility. However, since all tags can be moved, the only way to guarantee reproducibility is by pinning specific commit shas.

Likely it is easiest for us when using actions like these (e.g. checkout) to use the major version tag.

PR #590, opened 2022-12-06
ci: use major version for actions/setup-java

korthout:
## Description

Changes the version used for actions/setup-java in all workflows to major version `v3`.

This should reduce the amount of effort needed to update this action (i.e. no more dependabot pull requests unless a new major version is released).

It should also avoid problems when actions/setup-java removes a specific patch release tag (as has happened before, see pull request mentioned below).

Using a major version is possible because GH recommends moving the major release tags along with the minor and patch releases of that same version.

Note that this may result in reduced reproducibility. However, since all tags can be moved, the only way to guarantee reproducibility is by pinning specific commit shas. Since we don't necessarily depend on a specific version, we should trust the maintainers of the action to adhere to semver.

## Related issues

discussed in https://github.com/camunda/zeebe-process-test/pull/589#issuecomment-1338934871

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set up Java environment
-        uses: actions/setup-java@v3.6.0
+        uses: actions/setup-java@v3
         with:
           distribution: temurin
           java-version: 17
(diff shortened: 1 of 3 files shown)


saig0 (approved):
@korthout :+1: 

bors merge

PR #605, opened 2022-12-14
Inject ZPT related fields into superclasses

### LINKED ISSUE #604: @ZeebeProcessTest and injection is ignored when placed on a super class
There's a lot of boilerplate when it comes to activating and completing jobs. It would be ideal to place those common methods in a super class, annotated as ZeebeProcessTest and with client/engine field injection. However, in this case no field gets injected.

```java
@ZeebeProcessTest
public abstract class AbstractZeebeTest {

  protected ZeebeTestEngine engine;
  protected ZeebeClient client;

  @SneakyThrows
  public void waitForIdleState() {
    engine.waitForIdleState(ofSeconds(1));
  }

}
```

```java
public class PlanTest extends AbstractZeebeTest {

  @Test
  public void planTest() {
    waitForIdleState(); // engine will be null
  }

}
```

It would be great if this worked.

remcowesterhoud:
## Description

The extensions should inject the engine, client and recordstream into the fields that are available. Before this only tried to inject them into the annotated class. This does not work when inheritance is used and the fields are defined on the superclass of the annotated class.

This PR changes the extensions so that it will first look on the annotated class. If it cannot find any of the annotations here, it will work it's way up to the superclass and so on until it either found a field it is able to inject, or no fields at all.

## Related issues

closes #604 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/extension-testcontainer/src/main/java/io.camunda.zeebe.process.test.extension.testcontainer/ZeebeProcessTestExtension.java
+++ b/extension-testcontainer/src/main/java/io.camunda.zeebe.process.test.extension.testcontainer/ZeebeProcessTestExtension.java
@@ -123,14 +123,15 @@ public class ZeebeProcessTestExtension
 
   private void injectFields(final ExtensionContext extensionContext, final Object... objects) {
     final Class<?> requiredTestClass = extensionContext.getRequiredTestClass();
-    final Field[] declaredFields = requiredTestClass.getDeclaredFields();
     for (final Object object : objects) {
-      final Optional<Field> field = getField(declaredFields, object);
+      final Optional<Field> field = getField(requiredTestClass, object);
       field.ifPresent(value -> injectField(extensionContext, value, object));
     }
   }
 
-  private Optional<Field> getField(final Field[] declaredFields, final Object object) {
+  private Optional<Field> getField(final Class<?> requiredTestClass, final Object object) {
+    final Field[] declaredFields = requiredTestClass.getDeclaredFields();
+
     final List<Field> fields =
         Arrays.stream(declaredFields)
             .filter(field -> field.getType().isInstance(object))
@@ -143,8 +144,12 @@ public class ZeebeProcessTestExtension
                   + "found %s. Please make sure at most one field of type %s has been declared in the"
                   + " test class.",
               object.getClass().getSimpleName(), fields.size(), object.getClass().getSimpleName()));
+    } else if (fields.size() == 0) {
+      final Class<?> superclass = requiredTestClass.getSuperclass();
+      return superclass == null ? Optional.empty() : getField(superclass, object);
+    } else {
+      return Optional.of(fields.get(0));
     }
-    return fields.size() == 0 ? Optional.empty() : Optional.of(fields.get(0));
   }
 
   private void injectField(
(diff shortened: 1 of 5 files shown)


extension-testcontainer/src/main/java/io.camunda.zeebe.process.test.extension.testcontainer/ZeebeProcessTestExtension.java
```diff
@@ -143,8 +144,12 @@ private Optional<Field> getField(final Field[] declaredFields, final Object obje
                   + "found %s. Please make sure at most one field of type %s has been declared in the"
                   + " test class.",
               object.getClass().getSimpleName(), fields.size(), object.getClass().getSimpleName()));
+    } else if (fields.size() == 0) {
+      final Class<?> superclass = requiredTestClass.getSuperclass();
+      return superclass == null ? Optional.empty() : getField(superclass, object);
+    } else {
+      return Optional.of(fields.get(0));
```
koevskinikola, line 151: :+1: Nice, it will go up in the hierarchy until if finds something.

qa/embedded/src/test/java/io/camunda/zeebe/process/test/qa/embedded/injection/InheritanceInjectionTest.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.camunda.zeebe.process.test.qa.embedded.injection;
+
+import io.camunda.zeebe.process.test.extension.ZeebeProcessTest;
+import io.camunda.zeebe.process.test.qa.abstracts.injection.AbstractInheritanceInjectionTest;
+
+@ZeebeProcessTest
```
koevskinikola, line 22: :thought_balloon: I wonder if the `@ZeebeProcessTest` annotation should be placed on the `AbstractInheritanceInjectionTest`, as this is how it was reported in the issue. I tested it manually, and it works.

:thinking: On the other hand, it makes more sense to use the `@ZeebeProcessTest` annotation on the class where all the tests are.

:+1: In any case, the new code will work, and is properly tested.

koevskinikola (approved):
LGTM! :rocket: 

@remcowesterhoud thanks for passing the review to me. It was a nice chance to learn how the extension works.

I raised one point for discussion below, so this is why I didn't ask bors to merge the PR. It's optional, so fell free to merge it if you want to.

qa/embedded/src/test/java/io/camunda/zeebe/process/test/qa/embedded/injection/InheritanceInjectionTest.java
remcowesterhoud: Good question. I did think about this and decided against it in order to prevent duplication in our tests.

As you noticed we have 2 extensions in this project. One is for using an embedded Zeebe engine, the other is for using testcontainers. Depending on the Java version users can't use the embedded version, as it is tied to the Java version we use in Zeebe.

Because I didn't want to duplicate all testcases we have this "strange" test structure where we have all our tests defined in an abstract test class. Each of these abstract classes is inherited by 2 other test classes, 1 for the embedded extension and 1 for the testcontainers extension. The subclasses only provide the relevant extension to run the test using either the embedded engine, or the testcontainers engine.

As a result if we want to put the extension on the superclass, we would have to move the implementation of this test to `embedded` and the `testcontainers` module so we can test it for both extensions. As a result, we have a duplicated testcase 🙂

remcowesterhoud:
bors merge

qa/embedded/src/test/java/io/camunda/zeebe/process/test/qa/embedded/injection/InheritanceInjectionTest.java
koevskinikola: :+1: Then it makes sense to leave it as it is. The test is still correct, plus it avoids duplication.

PR #679, opened 2023-03-02
Apply recent API changes

Zelldon:
## Description
Return type is not necessary, since return was always true, which is why it has been removed from the API.

## Related issues

Related to changes made here https://github.com/camunda/zeebe/pull/11865/commits/b4e0e059587be3517606833208d85a0eebf80490

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseWriter.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseWriter.java
@@ -95,13 +95,13 @@ class GrpcResponseWriter implements CommandResponseWriter {
   }
 
   @Override
-  public boolean tryWriteResponse(final int requestStreamId, final long requestId) {
+  public void tryWriteResponse(final int requestStreamId, final long requestId) {
     if (rejectionType != RejectionType.NULL_VAL) {
       final Status rejectionResponse =
           responseMapper.createRejectionResponse(rejectionType, intent, rejectionReason);
       final Request request = gatewayRequestStore.removeRequest(requestId);
       sendError(request, rejectionResponse);
-      return true;
+      return;
     }
 
     try {
@@ -109,7 +109,6 @@ class GrpcResponseWriter implements CommandResponseWriter {
       final GeneratedMessageV3 response =
           responseMapper.map(request.requestType(), valueBufferView, key, intent);
       sendResponse(request, response);
-      return true;
     } catch (final Exception e) {
       throw new RuntimeException(e);
     }


remcowesterhoud (commented):
@Zelldon It seems there are some failing tests. I dug a bit into it and found out that there could be multiple `pendingResponses` for a single request.
image

This is unexpected to me, would you agree?

remcowesterhoud:
Maybe to explain why the tests are failing. We are keeping track of a map of requests

```
  Long registerNewRequest(
      final Class<? extends GeneratedMessageV3> requestType,
      final StreamObserver<?> responseObserver) {
    final long currentRequestId = requestIdGenerator.incrementAndGet();
    requestMap.put(currentRequestId, new Request(requestType, responseObserver));
    return currentRequestId;
  }
```

When a response is written we remove the request from this map and send the response to the client

```
  @Override
  public void tryWriteResponse(final int requestStreamId, final long requestId) {
    if (rejectionType != RejectionType.NULL_VAL) {
      final Status rejectionResponse =
          responseMapper.createRejectionResponse(rejectionType, intent, rejectionReason);
      final Request request = gatewayRequestStore.removeRequest(requestId);
      sendError(request, rejectionResponse);
      return;
    }

    try {
      final Request request = gatewayRequestStore.removeRequest(requestId);
      final GeneratedMessageV3 response =
          responseMapper.map(request.requestType(), valueBufferView, key, intent);
      sendResponse(request, response);
    } catch (final Exception e) {
      throw new RuntimeException(e);
    }
  }
```

Since there is multiple responses for a single request we run into an NPE. We could fix it by not sending a response if we can't find the corresponding request. But I'd like to make sure this isn't a bug in Zeebe.

Zelldon:
@remcowesterhoud I think it is a bug in zeebe. https://github.com/camunda/zeebe/blob/main/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java#L356

It is not really problematic, since the response is not sent twice if it was already sent, but yeah something which we should clean up.

I guess we should use a hashset instead of a list in the ProcessingStateMachine \cc @oleschoenburg

Zelldon:
Thanks for @remcowesterhoud checking the failures. I'm currently on FTO so have not time to fix it today.

oleschoenburg:
Fix is prepared: https://github.com/camunda/zeebe/pull/11892, I guess we can merge this tomorrow.

Zelldon:
Thanks to @oleschoenburg the CI is green again, I have rerun the jobs @remcowesterhoud I think we can merge it now.

remcowesterhoud (approved):
Perfect, thanks both!

bors merge

PR #686, opened 2023-03-09
fix(engine): fix compile issues in Engine creation

remcowesterhoud:
## Description

Some changes were made in Zeebe. The creation of the processors now requires an InterPartitionCommandSender directly, instead of a deploymentDistributionBehavior.

Besides this it also requires an EngineConfiguration.

## Related issues

closes #

## Definition of Done

Code changes:

Testing:

Documentation:


korthout:
@remcowesterhoud That seems to do the trick. The problem was that the Zeebe MessageState was altered once more and now sometimes uses `startAt` value of `null`. The DB needs to be able to deal with that.

korthout (approved):
LGTM 💯 

bors merge

PR #695, opened 2023-03-23
Fix dependency convergence issue

remcowesterhoud:
## Description

Fix spring dependency convergence issue which occurred [here](https://github.com/camunda/zeebe-process-test/actions/runs/4498336355/jobs/7914844261)


korthout (approved):
Thanks @remcowesterhoud 🚀 

LGTM 👍 

bors merge

PR #710, opened 2023-03-28
`ProcessInstanceAssert.hasCalledProcess` doesn't check if the process contains the passed process id

Wolgo:
Actually check if a process with the correct processID is present

## Description

Improve function to actually test for correct process ID. 

## Related issues

closes #

## Definition of Done

Code changes:

Testing:

Documentation:


Wolgo:
@remcowesterhoud I'm rudely tagging you, as I am otherwise going to forget about this, and it  Seems like a pretty important one, as this might result in tests incorrectly showing something calls the right subprocess.  Seems to me like this function was just forgetting to actually test for the correct ProcessId.

Didn't update tests or anything yet, because I have some other priorities right now.

remcowesterhoud (commented):
Thanks for the bugfix @Wolgo! Your changes look great, I like how you kept the old method available so we can stay backwards compatible 👍 

As you mentioned you haven't written tests. I can't merge this without tests, but I do want to see this merged. We can do 2 things here. If you are interested in writing the tests still you can commit this and I will merge this PR. Alternatively, I can create an issue and branch for this bug. I can merge it to this branch and write some tests for this myself. What do you prefer?

Wolgo:
`  public ProcessInstanceAssert findFirstCalledSubProcess(ProcessInstanceEvent ProcessEvent) {
    return new ProcessInstanceAssert(
        (StreamFilter.processInstance(this.recordStream)
            .withParentProcessInstanceKey(ProcessEvent.getProcessInstanceKey()).stream()
            .findFirst().orElseThrow(NoSuchElementException::new)).getKey(), this.recordStream);
  }`
  
  I quickly wrote this to work around that, and I'm thinking of just making it a generic function where you can get xth calls of process y.

remcowesterhoud:
> @remcowesterhoud I think it's good If I write some tests, do you have any examples? I saw these classes were not really doing any useful tests yet. I don't know if we wanna juggle with a mock setup or just use a real process.

It would be something along the lines of this in the `AbstractProcessInstanceAssertTest`:
```java
    @Test
    void testHasCalledProcess() throws InterruptedException, TimeoutException {
      // given
      Utilities.deployResources(
          client,
          ProcessPackCallActivity.RESOURCE_NAME,
          ProcessPackCallActivity.CALLED_RESOURCE_NAME);

      // when
      final ProcessInstanceEvent instanceEvent =
          Utilities.startProcessInstance(engine, client, ProcessPackCallActivity.PROCESS_ID);

      // then
      BpmnAssert.assertThat(instanceEvent)
          .hasCalledProcess(ProcessPackCallActivity.CALLED_PROCESS_ID);
    }
```

Of course this is only a happy path test for one of the functions. It's also not a great test as this is the only call activity that's in this process. So I would expect we need to introduce a `ProcessPackMultipleCallActivities` which contains a process with multiple call activities with different ids.

> I'm running to some other issues with subprocesses, so I'll probably extend it a bit more. (Multi-instances always seem to do X+1 calls, meaning that ExtractingLastCalledProcess becomes useless in the case of multi-instances.)

Would you mind separating this in a different branch? That way we can keep this PR isolated to one issue at a time.

remcowesterhoud:
Hi @Wolgo, hope you're doing well! This PR has been stale for a while, are you still interested in finishing this?

Wolgo:
Hey @remcowesterhoud , doing well, between a programming job and trying and succeeding a house, I never ended up finishing this.
Luckily you caught me on day where I could 0 coding and only did paperwork, so was pretty excited for this.
I wrote some tests for this, they are a bit wierd, as we can't check the processId of the current process in scope, so instead we use the fact that a process has or has not called a process to differentiate. (To me that's a bit easier to keep in mind, as that's already in scope, you could look at specific elements passed, but then you have to hold two concepts in mind.)

I am open for suggestions on alternatives that make more sense here. 
I added a small line about multi-instance loops, that's not a bug of this test code but maybe of camunda. Can't really do anything about that here, but it's a good warning for people :). 

I added a "HasNotCalledProcess" as that seems convenient in general and was really helpful for tests.

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/util/Utilities.java
```diff
@@ -244,7 +244,19 @@ public static final class ProcessPackTimerStartEvent {
...
     public static final String TIMER_ID = "timer";
   }
-
+  public static final class ProcessPackMultipleCallActivity {
+
+    public static final String RESOURCE_NAME = "multiple-call-activity.bpmn";
+    public static final String PROCESS_ID = "multiple-call-activity";
+    public static final String CALL_ACTIVITY_ID_ONE = "callactivityOne";
+    public static final String CALLED_RESOURCE_NAME_ONE = ProcessPackAlternateStartEndEvent.RESOURCE_NAME;
+    public static final String CALLED_PROCESS_ID_ONE = ProcessPackAlternateStartEndEvent.PROCESS_ID;
```
remcowesterhoud, line 253: ❌ Instead of introducing a `ProcessPackAlternateStartEndEvent` let's just use the `ProcessPackStartEndEvent`. This means you can reference the resource from the `ProcessPackStartEndEvent`. For the process id you'll need some unique string, e.g. `calledProcessIdOne`.

remcowesterhoud (changes requested):
> between a programming job and trying and succeeding to buy a house,

Sounds exciting!

Thanks for the changes you've made. Very thorough with the testing, I like it 🚀 

I have some minor suggestions. Mostly they are just some terminology fixes, but there's a few other changes as well. Please have a look and let me know if there's any questions!

Once you're done with the changes please run `mvn spotless:apply`. The CI is failing because of some code style issues 🙂

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/util/Utilities.java
Wolgo: I need a different processId between One and Three, to show a hasCalled(OneId) but hasNotCalled(ThreeId) on the parent project. 

Some of the requirements for my process that I have in mind: 
- Two sequential call activities, to show it correctly extracts the last called process.
- A call activity that has a call activity, to show the parent process does not detect nested child processes. (Which also means that the nested child process needs to have a different process Id than the 2 non-nested child processes.)

The process id is baked into the resource, so I made a new simple resource to prevent having to do some extra step.
Is there a way to deploy the same process twice with a different process id? 
Or is there another process that we can use for this?

Wolgo:
I'll finish the rest this evening, my pipeline is done, so back to actual work

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/util/Utilities.java
remcowesterhoud: Oh sorry, my bad. I got confused with how we create processes in the Zeebe repo. You are totally right that you need the second process and resource. You can disregard these comments about removing the resource.

qa/abstracts/src/main/resources/alternate-start-end.bpmn
```diff
@@ -0,0 +1,26 @@
...
+      <bpmndi:BPMNShape id="Event_0ahcnkn_di" bpmnElement="Event_0ahcnkn">
+        <dc:Bounds x="272" y="79" width="36" height="36" />
+      </bpmndi:BPMNShape>
+      <bpmndi:BPMNEdge id="Flow_0puuckb_di" bpmnElement="Flow_0puuckb">
+        <di:waypoint x="215" y="97" />
+        <di:waypoint x="272" y="97" />
+      </bpmndi:BPMNEdge>
+    </bpmndi:BPMNPlane>
+  </bpmndi:BPMNDiagram>
+</bpmn:definitions>
```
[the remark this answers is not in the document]
remcowesterhoud: Please ignore

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/util/Utilities.java
```diff
@@ -259,4 +271,9 @@ public static final class ProcessPackStartEndEvent {
     public static final String RESOURCE_NAME = "start-end.bpmn";
     public static final String PROCESS_ID = "start-end";
   }
+  public static final class ProcessPackAlternateStartEndEvent {
+
+    public static final String RESOURCE_NAME = "alternate-start-end.bpmn";
+    public static final String PROCESS_ID = "alternate-start-end";
+  }
```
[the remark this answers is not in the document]
remcowesterhoud: Please ignore

assertions/src/main/java/io/camunda/zeebe/process/test/assertions/ProcessInstanceAssert.java
```diff
@@ -593,18 +615,63 @@ public ProcessInstanceAssert extractingLatestCalledProcess(final String processI
...
+  /**
+   * Asserts whether this process has not called another process
+   *
+   * @return this {@link ProcessInstanceAssert}
+   */
+  public ProcessInstanceAssert hasNotCalledProcess() {
+    final boolean hasCalledProcess = getCalledProcessRecords().stream().findAny().isPresent();
+
+    assertThat(hasCalledProcess)
+        .withFailMessage("A process was called from this process")
```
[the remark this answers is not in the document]
Wolgo: I didn't see a way to count how many times each process is called, and if you are not intending to call a process, being told which distinct processes are called feels pretty good already.
But agreed, so implemented.

assertions/src/main/java/io/camunda/zeebe/process/test/assertions/ProcessInstanceAssert.java
```diff
@@ -634,7 +634,9 @@ public ProcessInstanceAssert hasNotCalledProcess() {
     final boolean hasCalledProcess = getCalledProcessRecords().stream().findAny().isPresent();
 
     assertThat(hasCalledProcess)
-        .withFailMessage("A process was called from this process")
+        .withFailMessage("A process was called from this process, distinct called processes are: %s",
+            getCalledProcessRecords().stream().map(
+                x -> x.getValue().getBpmnProcessId()).distinct().collect(Collectors.toList()).toString())
```
remcowesterhoud: 🔧 One more minor change before I can merge it. We already get the caller process records on line 634. Please store the result of this in a variable instead of getting them again here. E.g.:

```java
  public ProcessInstanceAssert hasNotCalledProcess() {
    final Set<String> calledProcesses =
        getCalledProcessRecords().stream()
            .map(process -> process.getValue().getBpmnProcessId())
            .collect(Collectors.toSet());

    assertThat(!calledProcesses.isEmpty())
        .withFailMessage(
            "A process was called from this process, distinct called processes are: %s",
            calledProcesses)
        .isFalse();
    return this;
  }
```

You will have to swap the processes in the failure message of your unit test.

remcowesterhoud (changes requested):
🚀 Great job, there is 1 more minor thing. After this I can merge the PR 🎉

assertions/src/main/java/io/camunda/zeebe/process/test/assertions/ProcessInstanceAssert.java
Wolgo: Ah, Collectors.toSet() removes the need for distinct. I still gotta get more used to streams in Java. 
Thanks for the reviewing!

remcowesterhoud (approved):
All good! Thanks for the contribution 🏆 And good luck with the new job and house 🙂

PR #768, opened 2023-05-16
ci(GHA): Migrate set-output to new GHA environment files

### LINKED ISSUE #669: Replace deprecated `::set-output` from CI (due date May 31st)
**Must be done before May 31st!**

**Description**

GitHub has deprecated the `::set-output` command. We are relying on this in our CI still. We must change this to the new standard.

How the change would look like to conform to the new standard:
before:
`echo "::set-output name=variable::value"`
after:
`echo "variable=value" >> "$GITHUB_OUTPUT"`

megglos:
## Description

As `set-output` is deprecated, see https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ and will actually get disabled end of May.

## Related issues

https://camunda.slack.com/archives/C01H4NG9XDY/p1683111124388649?thread_ts=1683032927.278609&cid=C01H4NG9XDY

closes #669

### DIFF
--- a/.github/workflows/deploy-artifact.yml
+++ b/.github/workflows/deploy-artifact.yml
@@ -115,7 +115,7 @@ jobs:
             echo "branch=stable/${MAJOR_MINOR_VERSION}" >> $GITHUB_OUTPUT
             echo "Branch = stable/${MAJOR_MINOR_VERSION}"
           else
-            echo "branch=stable/${{ github.event.repository.default_branch }}" >> $GITHUB_OUTPUT
+            echo "branch=${{ github.event.repository.default_branch }}" >> $GITHUB_OUTPUT
             echo "Branch = ${{ github.event.repository.default_branch }}"
           fi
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-artifact.yml
```diff
@@ -112,10 +112,10 @@ jobs:
           if ! [ -z "${{ github.event.release.tag_name }}" ] && [[ "${{ github.event.release.tag_name }}" != *"alpha"* ]]; then
             MAJOR_MINOR_VERSION=$(echo ${{ github.event.release.tag_name }} | sed -rn 's/^([0-9]+.[0-9]+).[0-9]+.*$/\1/p')
             test -z "${MAJOR_MINOR_VERSION}" && echo "::error::Tag ${{ github.event.release.tag_name }} does not adhere to semantic versioning" && exit 1
-            echo "::set-output name=branch::stable/$MAJOR_MINOR_VERSION"
-            echo "Branch = stable/$MAJOR_MINOR_VERSION"
+            echo "branch=stable/${MAJOR_MINOR_VERSION}" >> $GITHUB_OUTPUT
+            echo "Branch = stable/${MAJOR_MINOR_VERSION}"
           else
-            echo "::set-output name=branch::${{ github.event.repository.default_branch }}"
+            echo "branch=stable/${{ github.event.repository.default_branch }}" >> $GITHUB_OUTPUT
```
koevskinikola, line 118: :x: @megglos I don't think we need the `stable` prefix here since it is related to pushes to the `main` branch, or for `alpha` releases (as per the explanation in the code comment above).

```suggestion
            echo "branch=${{ github.event.repository.default_branch }}" >> $GITHUB_OUTPUT
```

koevskinikola (changes requested):

.github/workflows/deploy-artifact.yml
megglos: great catch, thanks!

koevskinikola (approved):
LGTM :rocket: 

bors merge

PR #803, opened 2023-06-16
[Backport stable/8.0] Verify processId when extracting called process

remcowesterhoud:
## Description

Extracting a called process takes a processId. This processId was used to verify the process has called a child process with this specific id. However, upon extracting the processId was ignored. As a result there was no guarantee that the extracted process actually matched the passed processId.

## Related issues

backport of #710


remcowesterhoud:
bors merge

PR #813, opened 2023-06-29
Added process definition inspections

jonathanlukas:
## Description

Process definition inspections were added.

## Definition of Done

Code changes:

Testing:

Documentation:


assertions/src/main/java/io/camunda/zeebe/process/test/inspections/InspectionUtility.java
```diff
@@ -33,4 +33,8 @@ public static ProcessInstanceInspections findProcessInstances() {
     return new ProcessInstanceInspections(
         StreamFilter.processInstance(BpmnAssert.getRecordStream()));
   }
+
+  public static DeploymentInspections findProcessDefinitions() {
+    return new DeploymentInspections(StreamFilter.deploymentRecords(BpmnAssert.getRecordStream()));
+  }
```
remcowesterhoud, line 39: ❌ Currently you're using the Deployment. I learned (after our chat yesterday) that we are removing the resources from the `Deployment.CREATED` event, in order to allow for larger deployments.

Instead, it's more clean to get the resource from the `Process.CREATED` event. Obviously this also requires some renaming of classes and splitting the DMN and BPMN parts. Personally I'd propose to leave DMN out of this PR as DMN isn't really supported in ZPT currently 😅

assertions/src/main/java/io/camunda/zeebe/process/test/inspections/DeploymentInspections.java
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Finds the BPMN element id by its name.
+   *
+   * <p>Keeps the test human-readable when asserting on bpmn elements. Asserts that there is only
+   * one element for the given name to prevent mistakes.
+   *
+   * @param elementName the name of the BPMN element
+   * @return the id of the found BPMN element
+   */
+  public String findBpmnElementId(String elementName) {
```
remcowesterhoud, line 52:
```suggestion
  public String getBpmnElementId(String elementName) {
```

🔧 Generally find is used when something returns an `Optional`, or in the earlier days the `elementId` or `null`. As this method will fail if it doesn't exist it's cleaner to change this to `get`. Please adjust the javadoc as well 🙂

assertions/src/main/java/io/camunda/zeebe/process/test/inspections/DeploymentInspections.java
```diff
@@ -0,0 +1,96 @@
...
+   * Finds the BPMN element id by its name inside the referenced BPMN process.
+   *
+   * <p>Keeps the test human-readable when asserting on bpmn elements. Asserts that there is only
+   * one element for the given name to prevent mistakes.
+   *
+   * @param bpmnProcessId the id of the deployed process
+   * @param elementName the name of the BPMN element
+   * @return the id of the found BPMN in the given process
+   */
+  public String findBpmnElementId(String bpmnProcessId, String elementName) {
```
remcowesterhoud, line 66:
```suggestion
  public String getBpmnElementId(String bpmnProcessId, String elementName) {
```

🔧 Generally find is used when something returns an `Optional`, or in the earlier days the `elementId` or `null`. As this method will fail if it doesn't exist it's cleaner to change this to `get`. Please adjust the javadoc as well 🙂

assertions/src/main/java/io/camunda/zeebe/process/test/inspections/DeploymentInspections.java
```diff
@@ -0,0 +1,96 @@
...
+    return findElementId(
+        deploymentRecordStreamFilter.withBpmnProcessId(bpmnProcessId).getBpmnDeploymentResources(),
+        elementName);
+  }
+
+  private String findElementId(Stream<DeploymentResource> stream, String elementName) {
```
remcowesterhoud, line 72:
```suggestion
  private String getElementId(Stream<DeploymentResource> stream, String elementName) {
```

🔧 Generally find is used when something returns an `Optional`, or in the earlier days the `elementId` or `null`. As this method will fail if it doesn't exist it's cleaner to change this to `get`. Please adjust the javadoc as well 🙂

remcowesterhoud (changes requested):
Hey @jonathanlukas, thanks for PR 🏅 

I have some comments regarding the event you get the resources from. We should use a different one 😄 

❌ Additionally to the other comments, I am missing some test cases to verify this works as expected.

assertions/src/main/java/io/camunda/zeebe/process/test/inspections/DeploymentInspections.java
jonathanlukas: Fixed in https://github.com/camunda/zeebe-process-test/pull/813/commits/d139f72b73a5eabc4519b65967b7cec4e3d52d8f

assertions/src/main/java/io/camunda/zeebe/process/test/inspections/InspectionUtility.java
jonathanlukas: Fixed in https://github.com/camunda/zeebe-process-test/pull/813/commits/d139f72b73a5eabc4519b65967b7cec4e3d52d8f

assertions/src/main/java/io/camunda/zeebe/process/test/inspections/DeploymentInspections.java
jonathanlukas: Fixed in https://github.com/camunda/zeebe-process-test/pull/813/commits/741b41591b16536bd8db80f64ab703c69e6d321b

assertions/src/main/java/io/camunda/zeebe/process/test/inspections/DeploymentInspections.java
jonathanlukas: Fixed in https://github.com/camunda/zeebe-process-test/pull/813/commits/741b41591b16536bd8db80f64ab703c69e6d321b

assertions/src/main/java/io/camunda/zeebe/process/test/inspections/DeploymentInspections.java
jonathanlukas: Fixed in https://github.com/camunda/zeebe-process-test/pull/813/commits/741b41591b16536bd8db80f64ab703c69e6d321b

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/inspections/AbstractProcessDefinitionInspectionsTest.java
```diff
@@ -0,0 +1,72 @@
...
+import io.camunda.zeebe.process.test.qa.abstracts.util.Utilities;
+import io.camunda.zeebe.process.test.qa.abstracts.util.Utilities.ProcessPackNamedElements;
+import org.junit.jupiter.api.Test;
+
+public abstract class AbstractProcessDefinitionInspectionsTest {
+  @Test
+  void testFindStartEventIdByName() {
+    // given
+    Utilities.deployResource(getClient(), ProcessPackNamedElements.RESOURCE_NAME);
+    Utilities.deployResource(getClient(), ProcessPackNamedElements.RESOURCE_NAME);
```
remcowesterhoud, line 34: ❓ Why are we deploying the process twice?

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/inspections/AbstractProcessDefinitionInspectionsTest.java
```diff
@@ -0,0 +1,72 @@
...
+    // when
+    String startEventId =
+        InspectionUtility.findProcessDefinitions()
+            .getBpmnElementId(ProcessPackNamedElements.START_EVENT_NAME);
```
remcowesterhoud, line 38: ❓ Do you think it's good enough UX like this? You probably write tests using ZPT more than I do. I'd imaging it might be inconvenient to have to call `InspectionUtility.findProcessDefinitions()` every time you need an element id. Storing it in a variable doesn't work because the stream will be closed if you use it a 2nd time.

remcowesterhoud: I feel like it might be more user friendly if we store the processes in the `ProcessDefinitionsInspections`. Something like this:

```java
  private final Map<String, BpmnModelInstance> processDefinitionsById;

  public ProcessDefinitionInspections(
      final ProcessDefinitionRecordStreamFilter deploymentRecordStreamFilter) {
    processDefinitionsById =
        deploymentRecordStreamFilter.withIntent(ProcessIntent.CREATED).stream()
            .collect(
                Collectors.toMap(
                    record -> record.getValue().getBpmnProcessId(),
                    record ->
                        Bpmn.readModelFromStream(
                            new ByteArrayInputStream(record.getValue().getResource()))));
  }
```

Then we can use it in the test like:

```java
    final ProcessDefinitionInspections processDef = InspectionUtility.findProcessDefinitions();
    final String startEventId =
        processDef.getBpmnElementId(ProcessPackNamedElements.START_EVENT_NAME);
    final String endEventId = processDef.getBpmnElementId(ProcessPackNamedElements.END_EVENT_NAME);
```

To be honest even nicer, IMO, would be if the instead of having the `InspectionUtility.findProcessDefinitions()` return a `ProcessDefinitionInspections` having it return a `ProcessDefinitions` object instead. This could then contain the convenience methods such as finding an element by id, or finding an element by process id and id. Happy to discuss to f2f if it's easier.

remcowesterhoud (changes requested):
Thanks for the changes @jonathanlukas. I have some more suggestions, curious to hear your thoughts about them!

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/inspections/AbstractProcessDefinitionInspectionsTest.java
jonathanlukas: This asserts that the `distinct()` in the stream resolving the id from the name works accordingly.

The assumption is that on deploying a resource twice, you would get 2 results containing the same id for a name which would then become one entry.

jonathanlukas: Should we create an explicit test for this instead?

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/inspections/AbstractProcessDefinitionInspectionsTest.java
jonathanlukas: In fact, I would static import `InspectionUtility` which would allow me to call:
```java
final String startEventId = findProcesses().getBpmnElementId(ProcessPackNamedElements.START_EVENT_NAME);
```
Sorry for the static import stuff. I do not like it for main code as well, but for testing, it is quite handy.

Your proposed solution could potentially lead to unexpected behaviour regarding the order of loading the utility and executing commands with zeebe.

filters/src/main/java/io/camunda/zeebe/process/test/filters/ProcessDefinitionRecordStreamFilter.java
```diff
@@ -0,0 +1,51 @@
...
+
+package io.camunda.zeebe.process.test.filters;
+
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.value.deployment.Process;
+import java.util.Objects;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+public class ProcessDefinitionRecordStreamFilter {
```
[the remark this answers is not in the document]
jonathanlukas: Solved in https://github.com/camunda/zeebe-process-test/pull/813/commits/71d3f96936e2cbce37105745ea7edb30395ca636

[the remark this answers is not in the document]
jonathanlukas: Solved in https://github.com/camunda/zeebe-process-test/pull/813/commits/71d3f96936e2cbce37105745ea7edb30395ca636

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/inspections/AbstractProcessDefinitionInspectionsTest.java
remcowesterhoud: That's not quite how Zeebe works. Zeebe will notice that the resource hasn't changed and not actually deploy a 2nd version of the process. If you want to test this you'll need a second process which contains en element with the same element id.

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/inspections/AbstractProcessDefinitionInspectionsTest.java
remcowesterhoud: To me it feels odd having to call `findProcesses()` every time. I found my processes, why would I do it again? It reads very much like this could be stored in a variable and reused, but that would break.

> Your proposed solution could potentially lead to unexpected behaviour regarding the order of loading the utility and executing commands with zeebe.

Imo this makes more sense. If I find my process definitions and decide to do a deployment after that it makes sense to me that the process definitions I found before don't contain this newly deployed process.

jonathanlukas: The ideal solution would be to stick to the C7 test pattern and make a static call for the method directly:
```java

final String startEventId = getBpmnElementId(ProcessPackNamedElements.START_EVENT_NAME);
```
This would require some class containing this static method which does not fit the rest of the test api. The solution you propose would require a change to the implementation so that the filtering on the record stream is only applied on method call?

remcowesterhoud: > The solution you propose would require a change to the implementation so that the filtering on the record stream is only applied on method call?

That was my suggestion indeed.

> The ideal solution would be to stick to the C7 test pattern and make a static call for the method directly

We can totally go with the static method as well. The API as-is is not a holy grail. If it's a better UX to have a static method we should go for it 🤷 
We just have to figure out where to put it. IMO we can fit that in the `inspections` package as well. Maybe something like `ProcessDefinitionInspectionUtility` class or so. We have access to the record stream in a static context, so the should be simple enough.

remcowesterhoud: Hey @jonathanlukas! Just tagging you to make sure you saw my previous message 🙂

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/inspections/AbstractProcessDefinitionInspectionsTest.java
jonathanlukas: I adjusted this. There are 2 different resources deployed now

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/inspections/AbstractProcessDefinitionInspectionsTest.java
jonathanlukas: I saw your previous message, sorry I was quite busy on friday. I just pushed something according to your suggestions.

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/inspections/AbstractProcessInspectionsTest.java
```diff
@@ -33,9 +34,7 @@ void testFindStartEventIdByName() {
     Utilities.deployResource(getClient(), ProcessPackNamedElements.RESOURCE_NAME);
     Utilities.deployResource(getClient(), ProcessPackNamedElements.RESOURCE_NAME_V2);
     // when
-    String startEventId =
-        InspectionUtility.findProcesses()
-            .getBpmnElementId(ProcessPackNamedElements.START_EVENT_NAME);
+    String startEventId = getBpmnElementId(ProcessPackNamedElements.START_EVENT_NAME);
```
remcowesterhoud, line 37: 🚀 Love the amount of boilerplate code that's gone!

remcowesterhoud (approved):
Great work @jonathanlukas 🏅 Thanks a lot for the contribution 🙇 

bors merge

remcowesterhoud:
bors retry

remcowesterhoud:
bors retry

PR #906, opened 2023-09-25
Support form deployment

### LINKED ISSUE #887: Add support to return Form metadata on DeployResource gRPC response
**Description**

To be able to return the Form metadata as a part of `DeployResource` gRPC response, `GrpcResponseMapper.createDeployResourceResponse` should be expanded.

Blocked by https://github.com/camunda/zeebe/issues/14131

nicpuppa:
## Description

closes #887 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/assertions/src/main/java/io/camunda/zeebe/process/test/assertions/DeploymentAssert.java
+++ b/assertions/src/main/java/io/camunda/zeebe/process/test/assertions/DeploymentAssert.java
@@ -138,7 +138,7 @@ public class DeploymentAssert extends AbstractAssert<DeploymentAssert, Deploymen
             formId, matchingForm.size(), matchingForm)
         .hasSize(1);
 
-    return new FormAssert(matchingForm.get(0));
+    return new FormAssert(matchingForm.get(0), recordStream);
   }
 
   /**
@@ -161,6 +161,6 @@ public class DeploymentAssert extends AbstractAssert<DeploymentAssert, Deploymen
             resourceName, matchingForm.size(), matchingForm)
         .hasSize(1);
 
-    return new FormAssert(matchingForm.get(0));
+    return new FormAssert(matchingForm.get(0), recordStream);
   }
 }
--- a/assertions/src/main/java/io/camunda/zeebe/process/test/assertions/DeploymentAssert.java
+++ b/assertions/src/main/java/io/camunda/zeebe/process/test/assertions/DeploymentAssert.java
@@ -78,7 +78,7 @@ public class DeploymentAssert extends AbstractAssert<DeploymentAssert, Deploymen
    * Extracts the process with the given BPMN process ID
    *
    * @param bpmnProcessId BPMN process ID to look up
-   * @return this {@link JobAssert}
+   * @return this {@link ProcessAssert}
    */
   public ProcessAssert extractingProcessByBpmnProcessId(final String bpmnProcessId) {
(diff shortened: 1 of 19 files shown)


assertions/src/main/java/io/camunda/zeebe/process/test/assertions/DeploymentAssert.java
```diff
@@ -118,4 +119,48 @@ public ProcessAssert extractingProcessByResourceName(final String resourceName)
 
     return new ProcessAssert(matchingProcesses.get(0), recordStream);
   }
+
+  /**
+   * Extracts the form with the given form ID
+   *
+   * @param formId form ID to look up
+   * @return this {@link JobAssert}
```
berkaycanbc, line 127: ❌ JavaDoc return type of this and the method below are wrong. Should be `FormAssert`.

assertions/src/main/java/io/camunda/zeebe/process/test/assertions/FormAssert.java
```diff
@@ -0,0 +1,109 @@
...
+ *         client.newDeployCommand().addResourceFile(file).send().join();
+ *
+ * final FormAssert formAssert =
+ *         assertThat(deploymentEvent)
+ *             .extractingProcessByBpmnProcessId(PROCESS_ID);
+ * }</pre>
+ */
+public class FormAssert extends AbstractAssert<FormAssert, Form> {
+
+  private final RecordStream recordStream;
```
berkaycanbc, line 41: 🔧 This one is unused. It would be good if we add `isCreated` method to assert whether form created record is present in the record stream. Therefore, this field will have an effect as well. Also, we need to add tests for the form created case.

berkaycanbc (changes requested):
Great contribution to a separate repo! @nicpuppa 🚀 

I left two minor comments and a suggestion which I think is better to add. 👍

assertions/src/main/java/io/camunda/zeebe/process/test/assertions/FormAssert.java
nicpuppa: Thanks for the input. I changed a bit the implementation, I added a new class for correctly inspect the log stream

nicpuppa:
Hey @berkaycanbc please have another look

assertions/src/main/java/io/camunda/zeebe/process/test/inspections/FormInspectionsUtility.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class FormInspectionsUtility {
+
+  /**
+   * Finds if a form is created.
+   *
+   * @param formId the id of the form
+   * @return true if the id of the form is found in the stream, false otherwise
+   */
+  public static boolean isFormCreated(final String formId) {
```
berkaycanbc, line 35: 🔧 It looks like to me that methods that asserts something in the form of `hasSomething` or `isDone` should be located in the assertion class. utility classes used for retrieving something from the stream. I would be better if we move this method into `FormAssert` class.

berkaycanbc (changes requested):
@nicpuppa thanks for the update! 👍 

I requested a change for a method that I think moving it to another class conforms better with the current structure.

berkaycanbc (approved):
Great work! LGTM! 🚀

nicpuppa:
bors merge

Issue #984, opened 2023-11-17
PullRequestProcessTest.testRemindReviewer is flaky


nicpuppa:
The test is failing with:

```
Error:  io.camunda.zeebe.process.test.examples.PullRequestProcessTest.testRemindReviewer -- Time elapsed: 4.264 s <<< FAILURE!
org.opentest4j.AssertionFailedError: Unable to activate 1 jobs, because only 0 were activated.
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:38)
	at org.junit.jupiter.api.Assertions.fail(Assertions.java:138)
	at io.camunda.zeebe.process.test.examples.PullRequestProcessTest.completeServiceTasks(PullRequestProcessTest.java:280)
	at io.camunda.zeebe.process.test.examples.PullRequestProcessTest.completeServiceTask(PullRequestProcessTest.java:269)
	at io.camunda.zeebe.process.test.examples.PullRequestProcessTest.testRemindReviewer(PullRequestProcessTest.java:136)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```
	

[More logs here](https://github.com/camunda/zeebe-process-test/files/13392306/Unit.Test.Results.ubuntu-latest.zip)

ana-vinogradova-camunda:
Just happened to me:
```
Error:  Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 9.016 s <<< FAILURE! -- in io.camunda.zeebe.process.test.examples.PullRequestProcessTest
[zb-actors-0] INFO io.camunda.zeebe.logstreams.impl.log.Sequencer - Closing sequencer for writing
Error:  io.camunda.zeebe.process.test.examples.PullRequestProcessTest.testRemindReviewer -- Time elapsed: 6.121 s <<< FAILURE!
[zb-actors-0] INFO io.camunda.zeebe.logstreams - On closing logstream null close 2 readers
org.opentest4j.AssertionFailedError: Unable to activate 1 jobs, because only 0 were activated.
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:38)
	at org.junit.jupiter.api.Assertions.fail(Assertions.java:138)
[...]

PR #1012, opened 2024-01-02
Support using process instance migration from the zeebe client

### LINKED ISSUE #972: Support using MigrateProcessInstance RPC
## Description

With https://github.com/camunda/zeebe/issues/15119 we have added a new `MigrateProcessInstance` RPC to the gRPC protocol. We must define this endpoint in the ZPT engine.

- [ ] Implement the RPC in the `GrpcToLogStreamGateway`
    - [ ] Add a response mapper to the `GrpcResponseMapper`

Would be nice to add a test case that shows that we can migrate a process instance in ZPT and then continue the process instance in the new model.

Blocked by
- https://github.com/camunda/zeebe/issues/15119

korthout:
## Description

This adds support for the new `MigrateProcessInstance` RPC to ZPT, such that users can try migrating process instances from ZPT. This allows testing migrations before doing so in production.

> [!NOTE]
> This does not add assertions for the migration. That is out of scope.

Additionally, this fixes a test case that no longer added value. The test case was intended to ensure that we don't forget supporting RPCs in ZPT, but the implementation of the test did not function correctly anymore. Now it is able to correctly detect unsupported RPCs again, which [highlighted that `streamActivatedJobs` is not supported by ZPT](https://github.com/camunda/zeebe/issues/11231#issuecomment-1874025863).

## Related issues

closes #972

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseMapper.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseMapper.java
@@ -40,6 +40,8 @@ import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.FailJobRequest;
 import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.FailJobResponse;
 import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.FormMetadata;
 import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.MatchedDecisionRule;
+import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.MigrateProcessInstanceRequest;
+import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.MigrateProcessInstanceResponse;
 import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.ModifyProcessInstanceRequest;
 import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.ModifyProcessInstanceResponse;
 import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.ProcessMetadata;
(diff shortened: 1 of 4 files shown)


engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseMapper.java
```diff
@@ -294,6 +300,27 @@ private GeneratedMessageV3 createModifyProcessInstanceResponse() {
...
+        .map(
+            instruction ->
+                MappingInstruction.newBuilder()
+                    .setSourceElementId(instruction.getSourceElementId())
+                    .setTargetElementId(instruction.getTargetElementId())
+                    .build())
+        .forEach(migrationPlanBuilder::addMappingInstructions);
+    final var migrationPlan = migrationPlanBuilder.build();
+
+    return MigrateProcessInstanceRequest.newBuilder()
```
berkaycanbc, line 318: ❌❓In the original response, we do not return any data. Is this intended to return data here for some purpose? @korthout

berkaycanbc (changes requested):
You left the code better than you found, thanks @korthout! 🚀 

I only requested one possible change.

engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseMapper.java
korthout: Good point. I tried to be too fast, and missed that I used `*Request` rather than `*Response`. The request was asking for this data, so I added it without looking any further.

I've fixed it up to:
```java
private GeneratedMessageV3 createMigrateProcessInstanceResponse() {
  return ModifyProcessInstanceResponse.getDefaultInstance();
}
```

berkaycanbc (approved):
LGTM! 🚀

korthout:
bors merge

korthout:
Known [flaky test](https://github.com/camunda/zeebe-process-test/issues/984)

bors retry

korthout:
Same flaky test

bors retry

korthout:
/backport

I forgot to backport it to the release branch, which had already been cut.

PR #1020, opened 2024-01-03
Add support for StreamActivatedJobs RPC

npepinpe:
## Description

This PR adds support for the `StreamActivatedJobs` RPC with `zeebe-process-test`, including basic back pressure implementation.

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -77,6 +77,21 @@
     </dependency>
 
     <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-scheduler</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-stream-platform</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>proto-google-common-protos</artifactId>
     </dependency>
@@ -114,6 +129,11 @@
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-stub</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
     </dependency>
 
     <!-- Test Scope -->
@@ -186,10 +206,10 @@
               <goal>analyze-duplicate</goal>
             </goals>
             <configuration>
-              <!-- TODO remove this after version of
-              plugin is released which includes
-              https://github.com/apache/maven-dependency-plugin/pull/194 -->
-              <failOnWarning>false</failOnWarning>
+              <ignoredUnusedDeclaredDependencies>
+                <!-- runtime binding for tests -->
+                <dependency>org.slf4j:slf4j-simple</dependency>
+              </ignoredUnusedDeclaredDependencies>
             </configuration>
           </execution>
         </executions>
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcToLogStreamGateway.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcToLogStreamGateway.java
@@ -7,10 +7,14 @@
  */
 package io.camunda.zeebe.process.test.engine;
 
+import static io.camunda.zeebe.util.buffer.BufferUtil.wrapString;
+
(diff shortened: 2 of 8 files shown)


engine/pom.xml
```diff
@@ -76,6 +76,21 @@
...
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-stream-platform</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
+
```
npepinpe, line 93: :thought_balloon: These dependencies were used in the engine module but not declared since we'd disabled failing the plugin previously. This was removed since the plugin was fixed, and the missing deps were added.

engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcToLogStreamGateway.java
```diff
@@ -137,6 +155,26 @@ public void activateJobs(
...
+    final var jobActivationProperties = new JobActivationPropertiesImpl();
+    final var worker = wrapString(request.getWorker());
+    final var jobType = wrapString(request.getType());
+    final var serverObserver = (ServerCallStreamObserver<ActivatedJob>) responseObserver;
+    final var consumer = new GrpcJobConsumer(jobType, serverObserver);
+    jobActivationProperties
+        .setWorker(worker, 0, worker.capacity())
+        .setTimeout(request.getTimeout())
+        .setFetchVariables(request.getFetchVariableList().stream().map(StringValue::new).toList())
+        .setTenantIds(request.getTenantIdsList());
```
npepinpe, line 171: :thought_balloon: Even though we don't support tenants yet here, not adding anything here would prevent ever streaming jobs (since there would be no tenant with the job stream, and the engine at least expects a default tenant).

engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcToLogStreamGateway.java
```diff
@@ -642,4 +680,46 @@ private DecisionEvaluationRecord createDecisionEvaluationRecord(
...
+  private final class GrpcJobConsumer implements JobConsumer, AutoCloseable {
+    private final DirectBuffer jobType;
+    private final ServerCallStreamObserver<ActivatedJob> observer;
+    private final Executor executor;
+
+    private GrpcJobConsumer(
+        final DirectBuffer jobType, final ServerCallStreamObserver<ActivatedJob> observer) {
+      this.jobType = jobType;
+      this.observer = observer;
+      executor = MoreExecutors.newSequentialExecutor(ForkJoinPool.commonPool());
```
npepinpe, line 693: :thought_balloon: gRPC docs requires us to synchronize access to the observer, so we use a sequential executor here wrapping the common fork join pool. Locks were an alternative to avoid async code, but I wasn't sure if that would impact the in memory engine since I'm not familiar with how it's written.

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
npepinpe, line 1: :thought_balloon: I'm not super happy with the tests. Without the actuator to know when the stream is registered, we can't accurately know which jobs are streamed/yielded. But I think coarse tests are fine here...they're maybe just a little less maintainable :(

engine/src/test/java/io/camunda/zeebe/process/test/engine/GrpcToLogStreamGatewayTest.java
```diff
@@ -22,7 +22,7 @@
 
 class GrpcToLogStreamGatewayTest {
 
-  static final List<String> UNSUPPORTED_METHODS = List.of("streamActivatedJobs");
+  static final List<String> UNSUPPORTED_METHODS = List.of();
```
npepinpe, line 25: :thought_balloon: I figured we might as well keep the mechanism, even if we now support all methods.

pom.xml
```diff
@@ -236,6 +236,18 @@
...
+        <artifactId>zeebe-scheduler</artifactId>
+        <version>${dependency.zeebe.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>io.camunda</groupId>
+        <artifactId>zeebe-stream-platform</artifactId>
+        <version>${dependency.zeebe.version}</version>
+      </dependency>
+
```
npepinpe, line 250: :thought_balloon: I didn't add these, but I think since we'd disabled the warning we missed them being now required.

koevskinikola (approved):
LGTM :+1: 

bors merge

PR #1028, opened 2024-01-12
Timer events fail with specific values

### LINKED ISSUE #975: Timer events Fails with specific values.
**Describe the bug**

We have two similar support cases describing the same issue where having 2 Timer boundary events fail for some specific dates.

**Timer events that get registered to be trigger at some specific dates will not be activated.**

https://jira.camunda.com/browse/SUPPORT-18817
https://jira.camunda.com/browse/SUPPORT-18790

**To Reproduce**

To reproduce the bug we only need to set up a process with the following configuration:

image

Setting the first non-interrupting timer boundary event with 10 days and the second interrupting boundary event with 20 days, and then configuring the system day for October 19th 2023.

This will cause the 10D trigger event not to get activated.

Both support tickets provided working examples of the tests that failed, see the following:
[zeebe-test-timer-events.tar.gz](https://github.com/camunda/zeebe/files/12915904/zeebe-test-timer-events.tar.gz)
[camunda-8-test-increase-time.zip](https://github.com/camunda/zeebe/files/12915906/camunda-8-test-increase-time.zip)

**Expected behavior**

The Timer boundary event should be activated after the 10 days.

One of the clients also did some further digging and provided one possible explanation for the behavior, namely that it might have something to low-level data manipulation bug of the timestamps:
<details><summary>Explanation</summary>
<p>

It seems there's a bug in the Zeebe test engine where TimerEvents are not correctly scheduled if the system time is within a specific time range. We noticed the exceptional behaviour only after executing tests after 2023-10-05 around 15:45:00+02:00 when suddenly a Timer Intermediate Catch Event with a duration of PT1S was no longer triggered after being activated. After multiple days of debugging and digging through internal Zeebe code, it seems that the order in which TimerEvents are triggered is inconsistent with the implementor's intention. The engine schedules two Timer Event configured as durations, where one is long-living (P30D) and the other one is due very shortly after activation (PT1S). However, if both events are scheduled and the clock of the process engine is after 2023-10-05T15:45:00+02:00, the short-lived timer is never activated, and the process tests run into an assertion timeout. We've set up an example project where the error is reliably reproduced. In the attached file you will find a minimal process and one test class (ProcessTimerTest.java) which has two unit tests:

 

will_fail_with_causing_date sets the engine clock to the first known point in time which triggers the error and fails
succeeds_with_time_travel_to_heyday sets the engine clock to the last point in time we know to work and succeeds
 

We have the suspicion that the test engine has a bug in the InMemoryDB where scheduled due dates are inserted/evaluated in the wrong order into a column which causes the Zeebe engine library, which is responsible of triggering scheduled timers, to first find the long-lived due date and stop processing other timer events. This leads to the vague suspicion that a very low-level somewhere bug exists which may be related to the byte order in which the timer is inserted into the DB. For context: a Timer Intermediate Catch Event with a duration is evaluated to the target trigger date as a Unix epoch timestamp (in milliseconds) which is stored as an 8 byte long in the DueDateColumn. Then engine then retrieves (only!) the very first timer from the column - relying on the ascending order of timestamps - and checks if it's over due by comparing its value to the current timestamp. If so, it triggers a TRIGGER command to be issued to the state machine. If not, it will stop processing because of the assumption that the first dueDate retrieved from the column is always the lowest of all inserted values.

How ever the comparison for values within the due date column is implemented for the real or the test engine, it should return the smaller timestamp before the bigger one, which fails for dueDates after the magic point in time (which is before 15:50 GMT+2 at the mentioned day) and is successful if the time of the engine is before that. 

</p>
</details>

**Environment:**

- OS: Mac OS Ventura / Windows 10 / Linux
- Zeebe Version: 8.2.16 and  8.1.2
- Configuration:

ana-vinogradova-camunda:
## Description

Fix for [this bug](https://github.com/camunda/zeebe-process-test/issues/975)

The root cause of this bug was in `Bytes` class `compareTo(final Bytes other)` method.

Before we used to compare byte arrays tin the following way:

```
if (ourByte < otherByte) {
        return SMALLER;
    } else if (ourByte > otherByte) {
        return BIGGER;
    } // else { // = equals -> continue }
```

This was leading to incorrect date comparison as we were comparing signed bytes and sometimes the date that was supposed to come after some other date had a negative byte (ex. -97). 

For example:
`2023.10.10 15:50:00` date will look as follows when converted to bytes array: `[0, 0, 1, -117, 25...]`
`2023.11.5 15:50:00` date will look as follows when converted to bytes array: `[0, 0, 1, -117, -97...]`

When comparing these 2 dates we would previously get `2023.11.5 15:50:00` as a smaller date that would result in `2023.10.10 15:50:00` timer event to not be activated.

The fix:
 I am using Guava `UnsignedBytes.compare(ourByte, otherByte)` method instead as it compares the two specified byte values, treating them as unsigned values between 0 and 255 inclusive. For example, (byte) -127 is considered greater than (byte) 127 because it is seen as having the value of positive 129.

## Related issues

closes #975

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/db/Bytes.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/db/Bytes.java
@@ -7,6 +7,7 @@
  */
 package io.camunda.zeebe.process.test.engine.db;
 
+import com.google.common.primitives.UnsignedBytes;
 import io.camunda.zeebe.db.DbValue;
 import java.util.Arrays;
 import org.agrona.ExpandableArrayBuffer;
@@ -21,6 +22,11 @@ final class Bytes implements Comparable<Bytes> {
   }
 
   @Override
+  public int hashCode() {
+    return Arrays.hashCode(byteArray);
+  }
+
+  @Override
   public boolean equals(final Object o) {
     if (this == o) {
       return true;
@@ -32,11 +38,6 @@ final class Bytes implements Comparable<Bytes> {
     final Bytes bytes = (Bytes) o;
 
     return Arrays.equals(byteArray, bytes.byteArray);
-  }
-
-  @Override
-  public int hashCode() {
-    return Arrays.hashCode(byteArray);
   }
 
   @Override
@@ -56,9 +57,10 @@ final class Bytes implements Comparable<Bytes> {
       final byte ourByte = byteArray[i];
       final byte otherByte = otherByteArray[i];
 
-      if (ourByte < otherByte) {
+      final int compared = UnsignedBytes.compare(ourByte, otherByte);
+      if (compared < 0) {
         return SMALLER;
-      } else if (ourByte > otherByte) {
+      } else if (compared > 0) {
         return BIGGER;
       } // else { // = equals -> continue }
     }
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/db/Bytes.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/db/Bytes.java
@@ -7,7 +7,6 @@
  */
 package io.camunda.zeebe.process.test.engine.db;
 
-import com.google.common.primitives.UnsignedBytes;
 import io.camunda.zeebe.db.DbValue;
 import java.util.Arrays;
 import org.agrona.ExpandableArrayBuffer;
@@ -42,35 +41,7 @@ final class Bytes implements Comparable<Bytes> {
 
   @Override
   public int compareTo(final Bytes other) {
-
-    final int EQUAL = 0;
-    final int SMALLER = -1;
-    final int BIGGER = 1;
-
-    final byte[] otherByteArray = other.byteArray;
-
-    for (int i = 0; i < byteArray.length; i++) {
-      if (i >= otherByteArray.length) {
-        return BIGGER;
-      }
-
-      final byte ourByte = byteArray[i];
-      final byte otherByte = otherByteArray[i];
-
-      final int compared = UnsignedBytes.compare(ourByte, otherByte);
-      if (compared < 0) {
-        return SMALLER;
-      } else if (compared > 0) {
-        return BIGGER;
-      } // else { // = equals -> continue }
-    }
-
-    if (byteArray.length == otherByteArray.length) {
-      return EQUAL;
-    } else {
-      // the other must be a longer array then
-      return SMALLER;
-    }
+    return Arrays.compareUnsigned(byteArray, other.byteArray);
   }
 
   byte[] toBytes() {
--- a/engine/src/test/java/io/camunda/zeebe/process/test/engine/db/BytesTest.java
+++ b/engine/src/test/java/io/camunda/zeebe/process/test/engine/db/BytesTest.java
@@ -18,7 +18,7 @@ import org.junit.jupiter.api.Test;
 public class BytesTest {
 
   @Test
-  public void compareDifferentSizeSmaller() {
+  public void shouldYieldShorterByteArrayAsSmaller() {
     // given
     final byte[] shorterArray = new byte[10];
     final byte[] longerArray = new byte[11];
@@ -32,7 +32,7 @@ public class BytesTest {
   }
 
   @Test
-  public void compareDifferentSizeBigger() {
+  public void shouldYieldLongerByteArrayAsBigger() {
     // given
     final byte[] shorterArray = new byte[10];
     final byte[] longerArray = new byte[11];
@@ -46,7 +46,7 @@ public class BytesTest {
   }
 
   @Test
-  public void compareEqual() {
+  public void shouldCompareEqualArraysAsZero() {
     // given
     final OffsetDateTime date =
         OffsetDateTime.of(2023, 10, 5, 15, 50, 0, 0, ZoneOffset.of("+02:00"));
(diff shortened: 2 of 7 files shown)


engine/src/main/java/io/camunda/zeebe/process/test/engine/db/Bytes.java
```diff
@@ -56,9 +57,10 @@ public int compareTo(final Bytes other) {
       final byte ourByte = byteArray[i];
       final byte otherByte = otherByteArray[i];
 
-      if (ourByte < otherByte) {
+      final int compared = UnsignedBytes.compare(ourByte, otherByte);
```
berkaycanbc, line 60: 🔧 Instead of depending on the external library, we can convert values to `int` using Java's `Byte.toUnsignedInt(byte x);` and keep the original comparison. 👍

oleschoenburg: There's also `Arrays#compareUnsigned` which does exactly what we want and probably faster than what we can implement.

tmetzke (changes requested):
👍 I like the new test cases and that we fix this in a "forward" manner.
🔧 Let's replace the custom comparison logic in `Bytes` completely with a call to [Arrays#compareUnsigned](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/util/Arrays.html#compareUnsigned-byte:A-byte:A-). @oleschoenburg is right, this does exactly what we want and way more efficiently than what is currently implemented here. For example, it has way better early-exit conditions and is null-safe.

engine/src/main/java/io/camunda/zeebe/process/test/engine/db/Bytes.java
ana-vinogradova-camunda: Thank you for the suggestion, sounds great 👍🏻

engine/src/main/java/io/camunda/zeebe/process/test/engine/db/Bytes.java
```diff
@@ -34,41 +39,9 @@ public boolean equals(final Object o) {
...
-      } // else { // = equals -> continue }
-    }
-
-    if (byteArray.length == otherByteArray.length) {
-      return EQUAL;
-    } else {
-      // the other must be a longer array then
-      return SMALLER;
-    }
+    return Arrays.compareUnsigned(byteArray, other.byteArray);
```
tmetzke, line 44: This is so much more maintainable 👍

engine/src/test/java/io/camunda/zeebe/process/test/engine/db/BytesTest.java
```diff
@@ -0,0 +1,141 @@
...
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import org.agrona.ExpandableArrayBuffer;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class BytesTest {
+
+  @Test
+  public void compareDifferentSizeSmaller() {
```
tmetzke, line 21: 🔧 Most of the test cases we write use the `shouldXYZ` naming convention. Should we use it here and describe what the test should ensure?
```suggestion
  public void shouldYieldShorterByteArrayAsSmaller() {
```

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/jobs/AbstractTimerTest.java
```diff
@@ -0,0 +1,177 @@
...
+  private RecordStream recordStream;
+
+  @BeforeEach
+  void deployProcesses() {
+    final DeploymentEvent deploymentEvent = Utilities.deployResource(client, RESOURCE);
+    BpmnAssert.assertThat(deploymentEvent).containsProcessesByResourceName(RESOURCE);
+  }
+
+  @Test
+  void testDateContainingNegativeByte() throws Exception {
```
tmetzke, line 57: 🔧 The test case name could be more descriptive regarding what it ensures. The `shouldXYZ` naming usually helps to reflect that more easily. The same goes for the other test cases in this class.

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/jobs/AbstractTimerTest.java
```diff
@@ -0,0 +1,177 @@
...
+import java.time.Duration;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.Optional;
+import java.util.stream.StreamSupport;
+import org.awaitility.Awaitility;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public abstract class AbstractTimerTest {
```
tmetzke, line 41: 🔧 It looks like we always use the same code in each test case but with a different date. This sounds like a perfect candidate for a parameterized test class.

tmetzke (changes requested):
👍 The compare method is so much better now 🚀 
🔧 Some smaller suggestions regarding test naming and test class setup. The test cases themselves look good to me, good job ⭐

engine/src/test/java/io/camunda/zeebe/process/test/engine/db/BytesTest.java
ana-vinogradova-camunda: Yes, absolutely. Will make this change now.

tmetzke (approved):

PR #1040, opened 2024-01-19
Unflake tests with published messages

### LINKED ISSUE #984: PullRequestProcessTest.testRemindReviewer is flaky

korthout:
## Description

ZPT test cases that manipulate time should not buffer messages for a shorter period than they want to travel through time because the message expires. 

Most often, the message expiry is not the event that your test wants to travel through time for, but rather you're doing so to wait for a timer event to trigger. However, time travel in ZPT is only deterministic in the sense that it awaits busy and idle state after time manipulation. That includes message expiry.

I've changed all tests to default to publishing messages without a TTL, so they aren't buffered. This means they cannot expire later during time travel. Instead, messages published without a TTL expire immediately when published.

I've also updated the example test case that showcases both message publication and time travel in the same test case.

This resolves a flaky test.

## Related issues

closes #984

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/examples/src/test/java/io/camunda/zeebe/process/test/examples/PullRequestProcessTest.java
+++ b/examples/src/test/java/io/camunda/zeebe/process/test/examples/PullRequestProcessTest.java
@@ -225,12 +225,33 @@ public class PullRequestProcessTest {
   private PublishMessageResponse sendMessage(
       final String messageName, final String correlationKey, final Map<String, Object> variables)
       throws InterruptedException, TimeoutException {
+
+    /*
+     To avoid flaky tests, we recommend publishing messages without a time to live when using time
+     manipulation in the same test case. Alternatively, you could plan out the timings of your time
+     manipulation and the published message's expiry.
+
+     In these tests, we assume that a timer event has triggered after the {@code increaseTime}
+     method returns. However, this is not guaranteed if a time to live is set because the message
+     could expire. Depending on the time to live, the message can expire due to time manipulation.
+
+     The {@code increaseTime} method will return after waiting for the engine to become idle again.
+     However, message expiry can cause the engine to be busy followed by being idle again. So, the
+     increaseTime method can return before the timer event has triggered when a message expires
+     due to time manipulation. This can be the cause of a flaky test.
+
+     Note that by default, the time to live is set to 1 hour.
+     See {@code ZeebeClientBuilder#defaultTimeToLive}.
+    */
+    final Duration timeToLive = Duration.ZERO;
+
     final PublishMessageResponse response =
         client
             .newPublishMessageCommand()
             .messageName(messageName)
             .correlationKey(correlationKey)
             .variables(variables)
+            .timeToLive(timeToLive)
             .send()
             .join();
     waitForIdleState(Duration.ofSeconds(1));
--- a/qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/assertions/AbstractMessageAssertTest.java
+++ b/qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/assertions/AbstractMessageAssertTest.java
@@ -136,7 +136,12 @@ public abstract class AbstractMessageAssertTest {
       // when
       final PublishMessageResponse response =
           Utilities.sendMessage(
-              engine, client, ProcessPackMessageEvent.MESSAGE_NAME, CORRELATION_KEY);
+              engine,
+              client,
+              ProcessPackMessageEvent.MESSAGE_NAME,
+              CORRELATION_KEY,
+              Duration.ofMinutes(1),
+              Collections.emptyMap());
 
       // then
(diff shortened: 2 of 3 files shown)


korthout:
I've run the related flaky test #984 locally, noticing that it failed about every third run. After the fix, I ran it successfully dozens of times. I've also run the entire test suite several times to verify that I hadn't created any new flaky tests.

tmetzke:
@korthout, have you run ZPT test cases by any chance, using your changes?
If I get your changes right, this test case in ZPT needs to be adjusted because now the message expires directly and the test case fails: https://github.com/camunda/zeebe-process-test/blob/main/qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/assertions/AbstractMessageAssertTest.java#L269.

korthout:
> @korthout, have you run ZPT test cases by any chance, using your changes? If I get your changes right, this test case in ZPT needs to be adjusted because now the message expires directly and the test case fails: https://github.com/camunda/zeebe-process-test/blob/main/qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/assertions/AbstractMessageAssertTest.java#L269.

The `testHasExpiredFailure` test case expects the message to not expire after publishing it. It depended on the message being published with a TTL of 1 minute (in `Utilities.sendMessage`). However, as I'm changing the behavior of this utility method to publish messages without a TTL, the message immediately expires on publication. That's expected behavior. The test must be adjusted because it explicitly needs to publish a message with some TTL.

[I have run all ZPT tests multiple times](https://github.com/camunda/zeebe-process-test/pull/1040#issuecomment-1900837206):
>  I've also run the entire test suite several times to verify that I hadn't created any new flaky tests.

Does that make sense @tmetzke ?

examples/src/test/java/io/camunda/zeebe/process/test/examples/PullRequestProcessTest.java
```diff
@@ -225,12 +225,33 @@ private void waitForBusyState(final Duration duration)
...
+     could expire. Depending on the time to live, the message can expire due to time manipulation.
+
+     The {@code increaseTime} method will return after waiting for the engine to become idle again.
+     However, message expiry can cause the engine to be busy followed by being idle again. So, the
+     increaseTime method can return before the timer event has triggered when a message expires
+     due to time manipulation. This can be the cause of a flaky test.
+
+     Note that by default, the time to live is set to 1 hour.
+     See {@code ZeebeClientBuilder#defaultTimeToLive}.
+    */
```
tmetzke, line 245: 💭 I was wondering if it makes sense to have such a comment in the utility class as well. Either just like it is or in a shorter form. In the end, before/when using the utility method, it would be nice to be made aware of this. What do you think?

tmetzke (approved):
👍 A really cool change, @korthout, I like the simplicity of it and the effect it has, making things more straightforward and robust here at the same time.
💭 I added a thought that shouldn't block you from merging here so I'll approve this already from my end.

examples/src/test/java/io/camunda/zeebe/process/test/examples/PullRequestProcessTest.java
korthout: 👍 I don't think the entire comment is necessary, but it would help to add JavaDoc that it publishes a message with TTL 0 and that an overloaded method exists to set a specific TTL. Note that the Utilities class is internal test only (it exists in test resources and is not something our users can use).

korthout: @tmetzke I've added a commit, could you have another look?

korthout:
Still using bors here - another reason to move ZPT into the Zeebe repo 🌟 

bors merge

Issue #960, opened 2023-11-07
Flaky test: `AbstractMessageAssertTest$UnhappyPathTests.testHasNotExpiredFailure`


korthout:
**Description**

Test failed and then succeeded in CI run: https://github.com/camunda/zeebe-process-test/actions/runs/6787134291/job/18449291513?pr=959

```
Warning:  Flakes: 
Warning:  io.camunda.zeebe.process.test.qa.abstracts.assertions.AbstractMessageAssertTest$UnhappyPathTests.testHasNotExpiredFailure
Error:    Run 1: AbstractMessageAssertTest$UnhappyPathTests.testHasNotExpiredFailure:302 
Expecting code to raise a throwable.
[INFO]   Run 2: PASS
```

AssertionError:
```
Error:  Tests run: 8, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 2.753 s <<< FAILURE! -- in io.camunda.zeebe.process.test.qa.abstracts.assertions.AbstractMessageAssertTest$UnhappyPathTests
[zb-actors-0] INFO io.camunda.zeebe.logstreams.impl.log.Sequencer - Closing sequencer for writing
Error:  io.camunda.zeebe.process.test.qa.abstracts.assertions.AbstractMessageAssertTest$UnhappyPathTests.testHasNotExpiredFailure -- Time elapsed: 0.657 s <<< FAILURE!
[zb-actors-0] INFO io.camunda.zeebe.logstreams - On closing logstream null close 2 readers
java.lang.AssertionError: 

Expecting code to raise a throwable.
	at io.camunda.zeebe.process.test.qa.abstracts.assertions.AbstractMessageAssertTest$UnhappyPathTests.testHasNotExpiredFailure(AbstractMessageAssertTest.java:302)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

I can't find the logs in the [`Unit Test Results macos-latest`](https://github.com/camunda/zeebe-process-test/suites/17983289905/artifacts/1034314325).

korthout:
@ana-vinogradova-camunda Please have a look, but time box it. It's okay if we're unable to resolve this now.

nicpuppa:
ZPA Mob-Programming:
[...]

PR #1056, opened 2024-01-31
Adding logging for further troubleshooting of a flaky test

ana-vinogradova-camunda:
Adding logging for further troubleshooting of a flaky test

## Description

Please see [this issue](https://github.com/camunda/zeebe-process-test/issues/960) for more detailed description.

I was unable to reproduce the issue and would like to suggest the following changes: 

- As Utilities.increaseTime() may swallow timeout exception if the engine takes more than 1 second to adjust to the time change and this might result in message expiration logic to be executed slightly later than expected resulting in assertion failure I suggest to log timeout exceptions separately to see on which state it failed
- As there might be the case when the message gets expired between assertion and test extension logging record stream, I suggest  logging record stream immediately before the assertion for further troubleshooting.

Please let me know your thought on this. 

## Related issues

closes #

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/assertions/AbstractMessageAssertTest.java
+++ b/qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/assertions/AbstractMessageAssertTest.java
@@ -22,6 +22,7 @@ import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
 import io.camunda.zeebe.client.api.response.PublishMessageResponse;
 import io.camunda.zeebe.process.test.api.ZeebeTestEngine;
 import io.camunda.zeebe.process.test.assertions.BpmnAssert;
+import io.camunda.zeebe.process.test.filters.RecordStream;
 import io.camunda.zeebe.process.test.qa.abstracts.util.Utilities;
 import io.camunda.zeebe.process.test.qa.abstracts.util.Utilities.ProcessPackMessageEvent;
 import io.camunda.zeebe.process.test.qa.abstracts.util.Utilities.ProcessPackMessageStartEvent;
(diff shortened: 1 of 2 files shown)


qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/util/Utilities.java
```diff
@@ -188,12 +192,24 @@ public static void increaseTime(final ZeebeTestEngine engine, final Duration dur
       throws InterruptedException {
     try {
       waitForIdleState(engine, Duration.ofSeconds(1));
-      engine.increaseTime(duration);
+    } catch (final TimeoutException e) {
```
berkaycanbc, line 195: 🔧 Let's add a detailed description why we added these different try catch errors to this method. Also, let's mention that these changes should be reverted when we understand the flakiness of the `UnhappyPathTests.testHasNotExpiredFailure`.

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/assertions/AbstractMessageAssertTest.java
```diff
@@ -308,6 +309,9 @@ void testHasNotExpiredFailure() throws InterruptedException, TimeoutException {
               Collections.emptyMap());
       Utilities.increaseTime(engine, timeToLive.plusMinutes(1));
 
+      // #960 logging current stream of events for assert troubleshooting
+      RecordStream.of(engine.getRecordStreamSource()).print(true);
```
berkaycanbc, line 313: ❌ Let's surround the assertion with try catch and log the record stream only if it fails. We can can re-throw the assertion error.

Please also consider adding a description that the code will be removed once the flakiness is resolved while adding the issue link.

berkaycanbc (changes requested):
Nice work @ana-vinogradova-camunda! 🚀 

I requested some changes that are important to add. 👍 

Additionally, I think, if these logs do not give much insight, we should consider asserting the time increase before continuing the test run.

qa/abstracts/src/main/java/io/camunda/zeebe/process/test/qa/abstracts/assertions/AbstractMessageAssertTest.java
ana-vinogradova-camunda: Thank you, great suggestion, done 👍🏻

berkaycanbc (approved):
@ana-vinogradova-camunda LGTM! 🚀

PR #1057, opened 2024-01-31
Fix InMemoryDb on zeebe 8.5.0-SNAPSHOT with changes to column family type

korthout:
## Description

The EnumValue is an interface that allows an enum to provide a getValue method that provides a number associated to each enum value. Hardcoding these values is less error prone than using the ordinal.

This was added to zeebe to avoid accidental reordering of column family values.

See:
- https://github.com/camunda/zeebe/pull/16106    

This also updates to 8.5.0-SNAPSHOT of Zeebe, which main should point at anyways.

## Related issues

https://github.com/camunda/zeebe-process-test/actions/runs/7720583417/job/21045722630

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/engine/src/test/java/io/camunda/zeebe/process/test/engine/db/InMemoryZeebeDbTransactionTest.java
+++ b/engine/src/test/java/io/camunda/zeebe/process/test/engine/db/InMemoryZeebeDbTransactionTest.java
@@ -15,6 +15,7 @@ import io.camunda.zeebe.db.TransactionContext;
 import io.camunda.zeebe.db.ZeebeDb;
 import io.camunda.zeebe.db.ZeebeDbTransaction;
 import io.camunda.zeebe.db.impl.DbLong;
+import io.camunda.zeebe.protocol.EnumValue;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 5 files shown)


engine/src/test/java/io/camunda/zeebe/process/test/engine/db/InMemoryZeebeDbTransactionTest.java
```diff
@@ -512,10 +513,21 @@ void shouldNotIterateOverDeletionsInTransaction() throws Exception {
         });
   }
 
-  private enum ColumnFamilies {
-    DEFAULT, // rocksDB needs a default column family
-    ONE,
-    TWO,
-    THREE
+  private enum ColumnFamilies implements EnumValue {
```
nicpuppa, line 516: ❓ Just a curiosity, why here we have an int value, and [here](https://github.com/camunda/zeebe/pull/16129/files#diff-f394f1f56b8a699fd006c23dd0a0b7519b05ae03b8097461def298c78350898fR480) instead we have the ordinal ?

nicpuppa (approved):
LGTM, I just left a question 🚀

engine/src/test/java/io/camunda/zeebe/process/test/engine/db/InMemoryZeebeDbTransactionTest.java
korthout: Because I was lazy [here](https://github.com/camunda/zeebe/pull/16129/files#diff-f394f1f56b8a699fd006c23dd0a0b7519b05ae03b8097461def298c78350898fR480). Static values are better 👍

korthout:
bors merge

PR #1087, opened 2024-02-13
1011 null pointer when sending out response with no request metadata

### LINKED ISSUE #1011: NullPointerException when sending out a response with no request metadata
## Description

The `GrpcResponseWriter` currently always expects a pending request for every response it sends out, and will explode with `NullPointerException` if this is not the case. Unfortunately, in the Zeebe engine, we sometimes send out responses anyway because the transport there simply ignores responses with no request metadata, i.e. request ID and stream ID are both the SBE null value, or `TypedRecord#hasRequestMetadata` returns false.

We should either mimic this behavior, or fix it on the Zeebe side. This currently prevents users from using any intermediate signal throw event.

### Expected behaviour

No NPEs are thrown, or the issue is fixed on the Zeebe side (no responses are sent without request metadata). It would likely be good to anyway catch it here.

## Reproduction steps

<details>
<summary>Test code</summary>

```
/*
 * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
 * one or more contributor license agreements. See the NOTICE file distributed
 * with this work for additional information regarding copyright ownership.
 * Licensed under the Zeebe Community License 1.1. You may not use this file
 * except in compliance with the Zeebe Community License 1.1.
 */

package io.camunda.zeebe.process.test.engine;

import io.camunda.zeebe.client.ZeebeClient;
import io.camunda.zeebe.model.bpmn.Bpmn;
import io.camunda.zeebe.process.test.api.ZeebeTestEngine;
import java.time.Duration;
import java.util.Map;
import java.util.concurrent.TimeoutException;
import org.junit.jupiter.api.AfterEach;
import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.extension.ExtendWith;

@ExtendWith(PrintRecordStreamExtension.class)
public class EngineTest {
  private ZeebeTestEngine zeebeEngine;
  private ZeebeClient zeebeClient;

  @BeforeEach
  void setupGrpcServer() {
    zeebeEngine = EngineFactory.create();
    zeebeEngine.start();
    zeebeClient = zeebeEngine.createClient();
  }

  @AfterEach
  void tearDown() {
    zeebeEngine.stop();
    zeebeClient.close();
  }

  // regression:
  @Test
  void shouldNotThrowNpeOnPublishMessageWhichBroadcastsSignal()
      throws InterruptedException, TimeoutException {
    // given
    final var process =
        Bpmn.createExecutableProcess("process")
            .startEvent()
            .intermediateCatchEvent(
                "start", b -> b.message(m -> m.name("start").zeebeCorrelationKey("=businessKey")))
            .intermediateThrowEvent("signal", b -> b.signal("signal"))
            .intermediateCatchEvent(
                "wait", b -> b.message(m -> m.name("wait").zeebeCorrelationKey("=businessKey")))
            .endEvent()
            .done();
    zeebeClient.newDeployResourceCommand().addProcessModel(process, "process.bpmn").send().join();
    final var pi =
        zeebeClient
            .newCreateInstanceCommand()
            .bpmnProcessId("process")
            .latestVersion()
            .variables(Map.of("businessKey", "42"))
            .send()
            .join()
            .getProcessInstanceKey();
    zeebeEngine.waitForIdleState(Duration.ofSeconds(600));

    // when
    zeebeClient.newPublishMessageCommand().messageName("start").correlationKey("42").send().join();
    zeebeEngine.waitForIdleState(Duration.ofSeconds(600));

    // then

  }
}

```

</details>

ana-vinogradova-camunda:
## Description

This PR is related to [this bug](https://github.com/camunda/zeebe/issues/15649). This was fixed and backported to 8.4 and 8.3 this week. 

The initial idea was to fix ZPT [GrpcResponseWriter](https://github.com/camunda/zeebe-process-test/blob/main/engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseWriter.java) `tryWriteResponse` method to contain similar check that was introduced in Zeebe project:

```
    if (command.hasRequestMetadata()) {
      responseWriter.writeEventOnCommand(eventKey, SignalIntent.BROADCASTED, signalRecord, command);
    }
```

More details are available in this comment [here](https://github.com/camunda/zeebe-process-test/issues/1011#issuecomment-1921617712). I am wondering if this is still valid as after the fix for [this bug](https://github.com/camunda/zeebe/issues/15649) the response for broadcast signal events that have no corresponding request is not sent anymore:

This check in `SignalBroadcastProcessor`:

```
    if (command.hasRequestMetadata()) {
      responseWriter.writeEventOnCommand(eventKey, SignalIntent.BROADCASTED, signalRecord, command);
    }
```

is happening before `GrpcResponseWriter` is called. 

I wonder if an extra check should still be added to `GrpcResponseWriter`? Do we have other scenarios when we don't have request metadata available? I will add the check if so 👍 

What is in this draft PR:
- Unit test that verifies that no response is sent for broadcast signal event that have no corresponding request 
- More details are available in separate commit messages

## Related issues

closes #1011 

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/EngineFactory.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/EngineFactory.java
@@ -99,7 +99,8 @@ public class EngineFactory {
         scheduler,
         recordStream,
         clock,
-        engineStateMonitor);
+        engineStateMonitor,
+        grpcResponseWriter);
   }
 
   private static ControlledActorClock createActorClock() {
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseWriter.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseWriter.java
@@ -19,6 +19,8 @@ import io.camunda.zeebe.util.buffer.BufferUtil;
 import io.camunda.zeebe.util.buffer.BufferWriter;
 import io.grpc.protobuf.StatusProto;
 import io.grpc.stub.StreamObserver;
+import java.util.ArrayList;
+import java.util.List;
 import org.agrona.DirectBuffer;
 import org.agrona.ExpandableArrayBuffer;
 import org.agrona.MutableDirectBuffer;
@@ -38,11 +40,16 @@ class GrpcResponseWriter implements CommandResponseWriter {
   private String rejectionReason = "";
   private final MutableDirectBuffer valueBuffer = new ExpandableArrayBuffer();
   private final GrpcResponseMapper responseMapper = new GrpcResponseMapper();
+  private final List<Intent> intents = new ArrayList<>();
 
   public GrpcResponseWriter(
       final GrpcToLogStreamGateway gateway, final GatewayRequestStore gatewayRequestStore) {
     this.gateway = gateway;
     this.gatewayRequestStore = gatewayRequestStore;
+  }
+
+  public List<Intent> getIntents() {
+    return intents;
   }
 
   @Override
@@ -109,6 +116,7 @@ class GrpcResponseWriter implements CommandResponseWriter {
       final GeneratedMessageV3 response =
           responseMapper.map(request.requestType(), valueBufferView, key, intent);
       sendResponse(request, response);
+      intents.add(intent);
     } catch (final Exception e) {
       throw new RuntimeException(e);
     }
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/InMemoryEngine.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/InMemoryEngine.java
@@ -41,6 +41,7 @@ public class InMemoryEngine implements ZeebeTestEngine {
   private final RecordStreamSource recordStream;
   private final ControlledActorClock clock;
   private final EngineStateMonitor engineStateMonitor;
+  private final GrpcResponseWriter grpcResponseWriter;
 
   public InMemoryEngine(
       final Server grpcServer,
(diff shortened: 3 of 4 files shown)


tmetzke (commented):
👍 I like the idea of testing this here to ensure it works after the fix in Zeebe.

ℹ️ Some general observations on this:
* We fixed the bug for signal processing specifically so far in Zeebe.
* We did **not** fix the general behavior of the response writer. This is fine for the initially [reported bug](https://github.com/camunda/zeebe/issues/15649) and this bug here.
* We can't reproduce the specific bug in ZPT anymore since we fixed it on command processor-level in Zeebe. This prevents it in ZPT as well.

💭 One question we should answer is the following: Do we want to fix this on the response writer-level as well?
* If so, let's do it in Zeebe and ZPT, I think (so the behavior is consistent).
* I don't have enough experience with Zeebe yet to state where this might happen as well, apart from signal processing. Might be wise to prevent this in the response writer in general if possible to save us from future headaches.

🔧 Regarding the PR specifically:
* Something like `grpcResponseWriter = ((InMemoryEngine) zeebeEngine).getGrpcResponseWriter();` might come back at us in the future. If we can generalize this a bit more, that would be great. Let me know if this does not make sense.
* Instead of collecting and returning the intents in the writer, use Mockito to verify the intent method has not been called with the intent you want to avoid. That's what you did in [this PR](https://github.com/camunda/zeebe/pull/15783/files#diff-b185102bc3dcf2db321c21433f41f3baf8792b7d61bdd858b30c4f7bb5e9862eR200) as well and it's the better choice, if possible.

ana-vinogradova-camunda:
Closing as decided on different implementation.

PR #1091, opened 2024-02-15
Test no response when no request metadata

### LINKED ISSUE #1011: NullPointerException when sending out a response with no request metadata

ana-vinogradova-camunda:
## Description

This PR is related to https://github.com/camunda/zeebe/issues/15649. This was fixed and backported to 8.4 and 8.3 this week.

Adding an optional request listener to `GrpcResponseWriter` `tryWriteResponse` method in order to verify the response is not send when there is no corresponding request.
So far we are only aware that this could happen with Intermediate Signal Throw events. 

If we have similar cases in the future we can address them then.
Please let me know if that's OK.

## Related issues

closes #1011

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseWriter.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseWriter.java
@@ -19,6 +19,7 @@ import io.camunda.zeebe.util.buffer.BufferUtil;
 import io.camunda.zeebe.util.buffer.BufferWriter;
 import io.grpc.protobuf.StatusProto;
 import io.grpc.stub.StreamObserver;
+import java.util.function.Consumer;
 import org.agrona.DirectBuffer;
 import org.agrona.ExpandableArrayBuffer;
 import org.agrona.MutableDirectBuffer;
@@ -38,11 +39,15 @@ class GrpcResponseWriter implements CommandResponseWriter {
   private String rejectionReason = "";
   private final MutableDirectBuffer valueBuffer = new ExpandableArrayBuffer();
   private final GrpcResponseMapper responseMapper = new GrpcResponseMapper();
+  private final Consumer<Intent> requestListener;
 
   public GrpcResponseWriter(
-      final GrpcToLogStreamGateway gateway, final GatewayRequestStore gatewayRequestStore) {
+      final GrpcToLogStreamGateway gateway,
+      final GatewayRequestStore gatewayRequestStore,
+      final Consumer<Intent> requestListener) {
     this.gateway = gateway;
     this.gatewayRequestStore = gatewayRequestStore;
+    this.requestListener = requestListener;
   }
 
   @Override
@@ -109,6 +114,9 @@ class GrpcResponseWriter implements CommandResponseWriter {
       final GeneratedMessageV3 response =
           responseMapper.map(request.requestType(), valueBufferView, key, intent);
       sendResponse(request, response);
+      if (requestListener != null) {
+        requestListener.accept(intent);
+      }
     } catch (final Exception e) {
       throw new RuntimeException(e);
     }
--- a/engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
+++ b/engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
@@ -9,6 +9,7 @@ package io.camunda.zeebe.process.test.engine;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 
 import com.google.common.util.concurrent.Uninterruptibles;
 import io.camunda.zeebe.client.ZeebeClient;
@@ -36,8 +37,10 @@ import io.camunda.zeebe.process.test.filters.JobRecordStreamFilter;
 import io.camunda.zeebe.process.test.filters.RecordStream;
 import io.camunda.zeebe.protocol.record.Record;
 import io.camunda.zeebe.protocol.record.RecordType;
+import io.camunda.zeebe.protocol.record.intent.Intent;
 import io.camunda.zeebe.protocol.record.intent.JobIntent;
 import io.camunda.zeebe.protocol.record.intent.ProcessInstanceIntent;
+import io.camunda.zeebe.protocol.record.intent.SignalIntent;
 import io.camunda.zeebe.protocol.record.intent.TimerIntent;
 import io.camunda.zeebe.protocol.record.value.BpmnElementType;
 import io.camunda.zeebe.protocol.record.value.BpmnEventType;
@@ -46,6 +49,7 @@ import io.camunda.zeebe.protocol.record.value.ProcessInstanceRecordValue;
 import io.camunda.zeebe.protocol.record.value.TimerRecordValue;
 import io.camunda.zeebe.util.VersionUtil;
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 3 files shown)


engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
```diff
@@ -1231,6 +1237,45 @@ void shouldBroadcastSignal() {
     assertThat(response.getKey()).isPositive();
   }
 
+  @Test
+  public void shouldNotSendResponseWhenThereNoCorrespondingRequest() {
```
tmetzke, line 1241: 🙃 We could go for either `shouldNotSendResponseWhenNoCorrespondingRequest` or `shouldNotSendResponseWhenThereIsNoCorrespondingRequest`. The `There` being there on its own reads a bit odd 🤷

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
```diff
@@ -1231,6 +1237,45 @@ void shouldBroadcastSignal() {
...
+                          false)
+                      .filter(r -> r.getValue().getBpmnElementType() == BpmnElementType.PROCESS)
+                      .filter(r -> r.getValue().getBpmnEventType() == BpmnEventType.UNSPECIFIED)
+                      .filter(r -> r.getIntent() == ProcessInstanceIntent.ELEMENT_COMPLETED)
+                      .findFirst();
+
+              assertThat(processCompleted).isNotEmpty();
+            });
+
+    assertFalse(responseIntents.contains(SignalIntent.BROADCASTED));
```
tmetzke, line 1276: 🔧 Let's not mix JUnit asserts and AssertJ. Rather use `assertThat(...).isFalse()` or assert the list directly, e.g. `assertThat(responseIntents).doesNotContain(SignalIntent.BROADCASTED)`.

tmetzke (approved):
👍 Changes look good, nice idea to track the intents along the way like that without storing them in the response writer.
❓ Have you tested this with Zeebe 8.4.x to see if it fails there?
🔧 I left some smaller suggestions, nothing that invalidates the solution so I approved it already.

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
ana-vinogradova-camunda: Totally agree, thank you 🙂

ana-vinogradova-camunda:
@tmetzke Zeebe fix related to that issue was back-ported to 8.4 and 8.3

tmetzke:
> @tmetzke Zeebe fix related to that issue was back-ported to 8.4 and 8.3

Fair enough, but there are versions of 8.4 before the backport you could try that test with, right? It's just another verification step to ensure we're testing something here that didn't work before and now works after the fix.

engine/src/main/java/io/camunda/zeebe/process/test/engine/GrpcResponseWriter.java
```diff
@@ -109,6 +114,9 @@ public void tryWriteResponse(final int requestStreamId, final long requestId) {
       final GeneratedMessageV3 response =
           responseMapper.map(request.requestType(), valueBufferView, key, intent);
       sendResponse(request, response);
+      if (requestListener != null) {
+        requestListener.accept(intent);
+      }
```
tmetzke, line 119: 💭 One thing that just came to my mind: what happens if the sendResponse fails? Do we still want to invoke the listener? Should we invoke it earlier? Otherwise, we won't know that it was invoked for the intent just because the method invocation failed. That might might lead to wrong assumptions, e.g. a response for the intent was not invoked although it was and we just didn't notice.

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
```diff
@@ -68,10 +71,12 @@ class EngineClientTest {
 
   private ZeebeTestEngine zeebeEngine;
   private ZeebeClient zeebeClient;
+  private List<Intent> responseIntents;
 
   @BeforeEach
   void setupGrpcServer() {
-    zeebeEngine = EngineFactory.create();
+    responseIntents = new ArrayList<>();
+    zeebeEngine = EngineFactory.create(responseIntents::add);
```
tmetzke, line 79: 🔧 We'll create the listener for each test case here. While the impact is probably low, we only might want to activate that listener for the new test case.

ana-vinogradova-camunda: 💭 I was looking at it, I am not sure how to do this in an easy and maintainable way tbh.

ana-vinogradova-camunda:
> > @tmetzke Zeebe fix related to that issue was back-ported to 8.4 and 8.3
> 
> Fair enough, but there are versions of 8.4 before the backport you could try that test with, right? It's just another verification step to ensure we're testing something here that didn't work before and now works after the fix.

Testing done, the test fails on versions before the backport.

engine/src/test/java/io/camunda/zeebe/process/test/engine/EngineClientTest.java
tmetzke: We could create the engine without the listener in the general `@BeforeEach` setup like before and only recreate it in the specific test case with the listener, I guess?

tmetzke (approved):
👍 Nice work!

PR #1178, opened 2024-05-07
fix: add context to CommandWriter

nicpuppa:
## Description

Added context to `CommandWrite` class due to the latest change in zeebe. See https://github.com/camunda/zeebe/pull/18266

### DIFF
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/CommandWriter.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/CommandWriter.java
@@ -10,6 +10,7 @@ package io.camunda.zeebe.process.test.engine;
 
 import io.camunda.zeebe.logstreams.log.LogAppendEntry;
 import io.camunda.zeebe.logstreams.log.LogStreamWriter;
+import io.camunda.zeebe.logstreams.log.WriteContext;
 import io.camunda.zeebe.protocol.impl.record.RecordMetadata;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/io/camunda/zeebe/process/test/engine/CommandWriter.java
```diff
@@ -22,14 +23,14 @@ record CommandWriter(LogStreamWriter writer) {
   void writeCommandWithKey(
       final Long key, final UnifiedRecordValue command, final RecordMetadata recordMetadata) {
     synchronized (writer) {
-      writer.tryWrite(LogAppendEntry.of(key, recordMetadata, command));
+      writer.tryWrite(WriteContext.userCommand(), LogAppendEntry.of(key, recordMetadata, command));
```
nicpuppa, line 26: ❓ Does it make sense to you @oleschoenburg ? I tried to understand the `WriteContext` usage

oleschoenburg: I think so :+1: If all writes are effectively user commands then it's perfect. Even if not, I don't think using the wrong `WriteContext` would matter for ZPT as it will only be used for flow control which we ignore here AFAIK.

nicpuppa:
Hey @oleschoenburg I addressed your comment, please have another look 👀

oleschoenburg:
Still looks good to me :)

PR #1182, opened 2024-05-14
fix: add highestPosition to logStorage

nicpuppa:
## Description

Added highestPosition to `InMemoryLogStorage` class due to the latest change in zeebe. See https://github.com/camunda/zeebe/pull/18297


rodrigo-lourenco-lopes (approved):
Thanks for the fix 🙌

PR #1205, opened 2024-07-17
fix: add message correlation log to record stream

nicpuppa:
## Description

Due to https://github.com/camunda/camunda/issues/20164 where a new Record Value is added


remcowesterhoud (approved):
Thanks @nicpuppa. You just gotta fix the code formatting 🙂

PR #1228, opened 2024-08-20
fix: build stream processor with SimpleMeterRegistry

nicpuppa:
## Description

Provide to stream processor a simple meter registry

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -136,6 +136,11 @@
       <artifactId>guava</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-core</artifactId>
+    </dependency>
+
     <!-- Test Scope -->
     <dependency>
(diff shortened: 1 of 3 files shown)


mustafadagher (commented):
The changes LGTM! But, let's check why the `TimerTest` is failing.

nicpuppa:
> The changes LGTM! But, let's check why the `TimerTest` is failing.

I already raised this in [this thread](https://camunda.slack.com/archives/C013MEVQ4M9/p1724123798977289). This PR fix only one issue with the tests @mustafadagher

korthout (approved):
👍

megglos:
@nicpuppa given this fixes the setup issue and the failure is an existing other issue, I would suggest to merge it to not have to include two different fixes into one PR. So once this is merged we can focus on the timer test fix separately, wdyt?

nicpuppa:
> Can we skip the queue for that?

I think we can disable the queue from the settings, let me check

Issue #1230, opened 2024-08-23
TimerTest started to fail consistently


ana-vinogradova-camunda:
**Description**

AbstractTimerTest was introduced [in this PR](https://github.com/camunda/zeebe-process-test/pull/1028) as a result of solving [this issue](https://github.com/camunda/zeebe-process-test/issues/975). 
The root cause of the issue was in Bytes class compareTo(final Bytes other) method.

Before we used to compare byte arrays tin the following way:
```
if (ourByte < otherByte) {
        return SMALLER;
    } else if (ourByte > otherByte) {
        return BIGGER;
    } // else { // = equals -> continue }
```
This was leading to incorrect date comparison as we were comparing signed bytes and sometimes the date that was supposed to come after some other date had a negative byte (ex. -97).

For example:
2023.10.10 15:50:00 date will look as follows when converted to bytes array: [0, 0, 1, -117, 25...]
2023.11.5 15:50:00 date will look as follows when converted to bytes array: [0, 0, 1, -117, -97...]

When comparing these 2 dates we would previously get 2023.11.5 15:50:00 as a smaller date that would result in 2023.10.10 15:50:00 timer event to not be activated.
The fix to use Guava UnsignedBytes.compare(ourByte, otherByte).

TimerTest failures were reported this week  (19 August) and the test is failing consistently now. 
I can reproduce the failure every time when running the test:

<details>
<summary><i>stack trace</i></summary>

```
[zb-actors-7] INFO io.camunda.zeebe.logstreams - Recovered state of partition 1 from snapshot at position -1
[zb-actors-7] INFO org.camunda.feel.FeelEngine - Engine created. [value-mapper: CompositeValueMapper(List(io.camunda.zeebe.feel.impl.MessagePackValueMapper@2d4b22a0)), function-provider: io.camunda.zeebe.feel.impl.FeelFunctionProvider@b981460, clock: io.camunda.zeebe.engine.processing.bpmn.clock.ZeebeFeelEngineClock@6e32ce2c, configuration: {externalFunctionsEnabled: false}]
[zb-actors-7] INFO org.camunda.dmn.DmnEngine - DMN-Engine created. [value-mapper: CompositeValueMapper(List(io.camunda.zeebe.feel.impl.MessagePackValueMapper@3efe5dec)), function-provider: org.camunda.feel.context.FunctionProvider$EmptyFunctionProvider$@59fb543a, audit-loggers: List(), configuration: Configuration(false,false,false)]
[...]

PR #1231, opened 2024-08-26
Add empty logger for authorization records

remcowesterhoud:
## Description

Authorizations aren't supported in the next version. The record is bound to change so it's not worth adding a proper logger for this value type right now. It likely isn't worth adding one in the future either, since ZPT is in the process of being deprecated.

## Related issues

N/A

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
+++ b/filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
@@ -100,17 +100,17 @@ public class RecordStreamLogger {
     valueTypeLoggers.put(ValueType.ESCALATION, this::logEscalationRecordValue);
 
     // These records don't have any interesting extra information for the user to log
-    valueTypeLoggers.put(ValueType.DEPLOYMENT_DISTRIBUTION, record -> "");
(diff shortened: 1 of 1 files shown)


filters/src/main/java/io/camunda/zeebe/process/test/filters/logger/RecordStreamLogger.java
```diff
@@ -131,6 +131,7 @@ public RecordStreamLogger(final RecordStreamSource recordStreamSource) {
     valueTypeLoggers.put(ValueType.MESSAGE_CORRELATION, this::logMessageCorrelationRecordValue);
     valueTypeLoggers.put(ValueType.USER, this::logUsersRecordValue);
     valueTypeLoggers.put(ValueType.CLOCK, this::logClockRecordValue);
+    valueTypeLoggers.put(ValueType.AUTHORIZATION, record -> "");
```
Zelldon, line 134: 🤯 Could we maybe like having a generic one with toString?

remcowesterhoud: That's fair, I'll change it for the other empty loggers too

remcowesterhoud:
CI fails because of https://github.com/camunda/zeebe-process-test/issues/1230 so we can't merge it yet until that one's fixed as well.

PR #1266, opened 2024-11-15
feat: add REDISTRIBUTION value type to logger

tmetzke:
## Description

Adds the REDISTRIBUTION value type to the record logger introduced with https://github.com/camunda/camunda/pull/24260.

## Related issues

related to https://github.com/camunda/camunda/issues/24013

## Definition of Done

Code changes:

Testing:

Documentation:


tmetzke:
@lenaschoenburg, if there is a specialized logging you'd consider appropriate here, feel free to make a proposal.

lenaschoenburg:
Oh thanks for implementing this! I don't thinks it's important how we log it because redistribution will never get triggered in ZPT anyway. So any format works for me.

lenaschoenburg (approved):

tmetzke:
> Oh thanks for implementing this! I don't thinks it's important how we log it because redistribution will never get triggered in ZPT anyway. So any format works for me.

That's what I assumed as well and why I went for `toString` 😄 👍

PR #1306, opened 2025-01-02
ci: revert nexus.snapshot.repository.id change

megglos:
## Description

The change of the id caused issues on artifact deploy, as it needs to match the id of the repo from the maven settings to use the correct credentials.

It previously broke because of https://github.com/camunda/zeebe-process-test/pull/1286


pom.xml
```diff
@@ -486,7 +486,7 @@
       <snapshots>
         <enabled>false</enabled>
       </snapshots>
-      <id>${nexus.release.repository.id}</id>
+      <id>${nexus.release.repository.id}-release</id>
```
megglos, line 489: as there wan't an entry for the release repo entry at all recently, I hope just changing the name here should not cause side effects 😬

npepinpe (approved):
Let's try it

PR #1388, opened 2025-01-05
build: run TC tests with Java 21

npepinpe:
## Description

We specify the release version to be Java 8 when running the tests, which ensures that we will catch issues where use APIs > JDK 8, or syntax that is unsupported on JDK 8, while still allowing us to use a build environment (including plugins) for greater versions.

The one downside is we won't detect anymore if we pull in transitive dependencies > JDK 8, but this should be detected separately.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -113,18 +113,9 @@ jobs:
           IMAGE_NAME_KEY: container.image.name
           IMAGE_TAG_KEY: container.image.tag
 
-      - name: Downgrade Java environment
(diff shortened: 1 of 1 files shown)


megglos:
@npepinpe what was the reason for this change, simplifying/speeding up the CI? I stumbled over it because it's still present on stable branches <=8.7 and I wonder whether we just remove it everywhere now 🤔

npepinpe:
IIRC it was causing some problems, but I can't remember anymore. Failure on my side, that should've been captured in the PR description.

PR #1494, opened 2025-02-17
deps: align camunda model version to 7.22 with camunda/camunda

megglos:
## Description

Quick fix for potential issues caused by the version misalignment. It's a bit unfortunate this needs manual alignment, on the other hand it's not the only dep that needs to stay aligned and this one is rather infrequent to change.

## Related issues

closes #1493 

## Definition of Done

_Not all items need to be done depending on the issue and the pull request._

Code changes:
* [ ] The changes are backwards compatibility with previous versions
* [ ] If it fixes a bug then PRs are created to backport the fix

Testing:
* [ ] There are unit/integration tests that verify all acceptance criterias of the issue
* [ ] New tests are written to ensure backwards compatibility with further versions
* [ ] The behavior is tested manually

Documentation:
* [ ] Javadoc has been written
* [ ] The documentation is updated

PR #1500, opened 2025-02-17
deps: use root pom version for module dependencies

megglos:
## Description

Module versions got messed up by renovate.

See e.g. https://github.com/camunda/zeebe-process-test/commit/805ddacc8b76a9ff199b810fa98b2b8e051d1b20 , we need to exclude the repo modules in renovate in a follow-up.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>io.camunda</groupId>
     <artifactId>zeebe-process-test-root</artifactId>
-    <version>8.6.9</version>
+    <version>8.6.10-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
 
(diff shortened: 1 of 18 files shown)


megglos:
actually the CI https://github.com/camunda/zeebe-process-test/pull/1494 shouldn't have passed as e.g. backports to 8.5 failed with a compile error, given that the model dependencies are not jdk8 compatible anymore.

I wondered why the 8.6 PR passed and found the module dependency setup to be broken, this fixes the setup to use the actual local modules from the maven reactor and as expected breaks the CI due to the jdk 8 conflict.

I would merge this like this still as this is the actual state of the branch and other builds have an inconsistent dependency setup.

megglos:
merging eagerly to make progress and get a reliable CI, happy to apply suggestion in a follow-up

PR #1557, opened 2025-03-21
Log all unknown records using toString

korthout:
## Description

Every time a record value is added, the RecordStreamLoggerTest fails because we need to implement a nice record logger for that record value.

However, every time we just implement it with toString, which internally uses toJson to fix the failing test.

Instead, of failing every time, we can just accept that new records are logged as json.


remcowesterhoud (approved):
Awesome 👏

PR #1558, opened 2025-03-24
Fix snapshot build

korthout:
## Description

- Ensures we log during tests
- Ensures we log all camunda logs at debug level
- Provides a meter registry bean if missing

## Related issues

https://camunda.slack.com/archives/C013MEVQ4M9/p1742786448401559

## Definition of Done

Code changes:

Testing:

Documentation:

### DIFF
--- a/assertions/src/main/resources/log4j2.xml
+++ b/assertions/src/main/resources/log4j2.xml
@@ -9,7 +9,7 @@
 
     <Loggers>
         <Logger name="io.zeebe" level="debug"/>
-        <Logger name="io.camunda.zeebe" level="debug"/>
+        <Logger name="io.camunda" level="debug"/>
 
         <Root level="info">
             <AppenderRef ref="Console"/>
--- a/engine/src/main/resources/log4j2.xml
+++ b/engine/src/main/resources/log4j2.xml
@@ -9,7 +9,7 @@
 
     <Loggers>
         <Logger name="io.zeebe" level="debug"/>
-        <Logger name="io.camunda.zeebe" level="debug"/>
+        <Logger name="io.camunda" level="debug"/>
 
         <Root level="info">
             <AppenderRef ref="Console"/>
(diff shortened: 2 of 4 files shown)


remcowesterhoud (approved):
Thanks @korthout!

megglos:
@korthout 
Following up on adding the MeterRegistry here, in https://github.com/camunda/zeebe-process-test/pull/1617#issuecomment-2877467613 I found this is only needed here because the actual default [NoopRegistry](https://github.com/camunda/camunda/blob/stable/8.6/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/zeebe/spring/client/configuration/MetricsDefaultConfiguration.java) is not loaded because we always have the `spring-boot-actuator-autoconfigure` dep on classpath when the spring sdk is used. Which seems like a mistake, as this way the NoopRegistry is never loaded unless you explicitly exclude the autoconfigure dep, like I did [here](https://github.com/camunda/zeebe-process-test/commit/f26795e35385bdce5887a9cddbdaf78d7fe17eee) to fix the test failures.

Thus my question, was there a reason to actually go with adding micrometer explicitly instead of going for the NoopRegistry?

megglos:
See https://github.com/camunda/camunda/pull/31991#issuecomment-2879648558 on the root-cause. I will create a follow-up PR to remove the MetricsRegistry setup as it's not needed for the ZPT test setup once these deps are not inherited.

korthout:
> was there a reason to actually go with adding micrometer explicitly instead of going for the NoopRegistry?

I went for SimpleMeterRegistry, as it was a straightforward solution that exists specifically for testing use cases which ZPT is. I believed that the impact would be small.

>Micrometer includes a SimpleMeterRegistry that holds the latest value of each meter in memory and does not export the data anywhere.

However, I agree that a NoopRegistry would be better because we have no need for metrics in ZPT.

PR #1603, opened 2025-04-29
test: fix test relying availability of implicit test bean

megglos:
## Description

The test setup was relying on the presence of an implicit bean representing the test, being `io.camunda.zeebe.spring.client.jobhandling.JobHandlerTest.ORIGINAL`. Which is not present in the `applicationContext` used now to process worker annotations The fix moves all worker setups into dedicated configuration classes to resolve this.

The cause for it breaking now is a change to SDK but this is nothing relevant for the actual use of the SDK in production code, where workers would always be part of the explicit spring configuration.

See https://camunda.slack.com/archives/C06UWQNCU7M/p1745920644082269

### DIFF
--- a/spring-test/embedded/src/test/java/io/camunda/zeebe/spring/client/jobhandling/JobHandlerTest.java
+++ b/spring-test/embedded/src/test/java/io/camunda/zeebe/spring/client/jobhandling/JobHandlerTest.java
@@ -52,6 +52,7 @@ import org.springframework.context.annotation.Bean;
 @SpringBootTest(
     classes = {
       JobHandlerTest.class,
+      JobHandlerTest.WorkerConfig.class,
       JobHandlerTest.TestMetricsConfiguration.class,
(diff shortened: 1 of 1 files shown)


jonathanlukas (approved):
Thank you for fixing that!

megglos:
For reference the causing change was https://github.com/camunda/camunda/pull/30820 which moved the worker setup from being run [during PostProcessing](https://github.com/camunda/camunda/pull/30820/files#diff-6732aa614ba708308bef0316ade2a5a2136a106c76fa8403b2c4f2a32061f084L38) to act on to [Camunda Client LifeCycle Events only](https://github.com/camunda/camunda/pull/30820/files#diff-1bc0b16c20fc3449c222cdaff3718cc678930398c7b2750bc1d21fc04abf7da6R41).

In that case the implicitly created bean representing the test class is not available in the `applicationContext`.

PR #1607, opened 2025-05-05
test: fix smoketest relying on implicit test bean

megglos:
## Description

Same fix as https://github.com/camunda/zeebe-process-test/pull/1603 for another test.

The test setup was relying on the presence of an implicit bean representing the test. Which is not exposed in the `applicationContext` by default. The fix moves all worker setups into dedicated configuration classes to resolve this.


jonathanlukas (approved):
Cool! Also good catch with the properties mapping!

megglos:
> Cool! Also good catch with the properties mapping!

All props go to IntelliJ hinting me on the usage of deprecated config :D

Issue #1619, opened 2025-05-07
Unstable MacOS CI


megglos:
**Description**

Issues like https://github.com/camunda/zeebe-process-test/issues/960 and also more recent observations during the release see https://camunda.slack.com/archives/C08F5RD5X8B/p1746529769683449 indicate general instability of the [MacOS CI](https://github.com/camunda/zeebe-process-test/blob/main/.github/workflows/build-test.yml) in the `build-test` workflow.

This impacts an efficient release we thus either resolve the instability or revisit whether it's worth to run that stage.

cmur2:
This also affected the time critical 8.8.0-alpha4.1 hotfix release.

@megglos Please prioritize addressing this instability short-term e.g. by ignoring the stage.

A long-term fix is then less time-critical.

PR #1620, opened 2025-05-07
ci: collect surefire reports from nested modules

megglos:
## Description

Previous setup was just collecting the report from the top level modules, due to the non-recursive glob pattern used.

relates https://github.com/camunda/zeebe-process-test/issues/1619 - for being able to check the logs of the flaky tests


saig0 (approved):
:+1:

PR #1624, opened 2025-05-09
fix: fix method signature in noop search client to match upstream ver…

stoerti:
## Description

This fixes a compile error coming from the upstream Camunda SNAPSHOT version. Where the `long batchOperationKey` was renamed to the `String batchOperationId` in the `BatchOperationSearchClient` interface

## Related issues

https://github.com/camunda/camunda/issues/31589

### DIFF
--- a/engine/src/main/java/io/camunda/zeebe/process/test/engine/NoResultsSearchClientsProxy.java
+++ b/engine/src/main/java/io/camunda/zeebe/process/test/engine/NoResultsSearchClientsProxy.java
@@ -81,7 +81,7 @@ class NoResultsSearchClientsProxy implements SearchClientsProxy {
   }
 
   @Override
-  public List<BatchOperationItemEntity> getBatchOperationItems(final Long batchOperationKey) {
+  public List<BatchOperationItemEntity> getBatchOperationItems(final String batchOperationKey) {
     return List.of();
   }
 


tmetzke:
Test run of the Zeebe SNAPSHOT with the fix: https://github.com/camunda/zeebe-process-test/actions/runs/14922589209

tmetzke:
> Test run of the Zeebe SNAPSHOT with the fix: https://github.com/camunda/zeebe-process-test/actions/runs/14922589209

Interesting, that run now fails with what we saw yesterday already:

```
Error:  io.camunda.zeebe.process.test.qa.abstracts.assertions.AbstractIncidentAssertTest$HappyPathTests.testIsUnresolved -- Time elapsed: 0.267 s <<< ERROR!
java.lang.RuntimeException: 
com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `io.camunda.zeebe.protocol.record.value.AuthorizationResourceType` from String "BATCH": not one of the values accepted for Enum class: [BATCH_OPERATION, PROCESS_DEFINITION, MESSAGE, TENANT, USER, MAPPING_RULE, RESOURCE, SYSTEM, APPLICATION, ROLE, UNSPECIFIED, AUTHORIZATION, DECISION_REQUIREMENTS_DEFINITION, DECISION_DEFINITION, GROUP]
```

Do we need to adjust that as well or is the action run using an old SNAPSHOT of Zeebe?

megglos (approved):
Jumping in here as @camunda/camundaex is taking ownership of this repo. As it resolved a compile error would like to merge it. Will follow-up on the potentially still instable CI separately.

megglos:
Thank you @stoerti !

Issue #1622, opened 2025-05-07
Renovate updates mess up multi-module versioning


megglos:
## Description

Taking e.g. stable/8.6 as an example https://github.com/camunda/zeebe-process-test/blob/stable/8.6/pom.xml#L137-L177

Renovate has messed up the version references to other modules in the multi-module build.
This likely affects all branches and may lead to unexpected behavior, given some modules make use of previous releases of other modules.

### Expected behaviour

Ideally renovate is not updating any of the versioning of modules part of the multi-module project. We either get rid of the explicit versioning or setup renovate to ignore it.

PR #1629, opened 2025-05-09
ci: exclude io.camunda.** from renovate updates

megglos:
## Description

It messed up inter-module dependencies (see #1622) and we generally maintain dependencies to camunda/camunda manually as part of the release.

## Related issues

relates #1622

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -86,6 +86,16 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


megglos:
merging eagerly to stop renovate from doing these updates asap. If you have feedback/different suggestion happy to follow-up on that.

PR #1617, opened 2025-05-07
[backport stable/8.7] ci: run spring testcontainers test in tc job

megglos:
## Description

backport of https://github.com/camunda/zeebe-process-test/pull/1609

With additional changes needed to account for the state of stable/8.7

1.  Backport of https://github.com/camunda/zeebe-process-test/pull/1388 to remove jdk downgrade, that would cause this change to fail given that the sdk tc job requires jdk 17, done via 0e8632fa52fd5fb6d1ae6c1d400a268969ef594c
2. Adding slf4j-simple via b02acc93c8eaa55825fbfd8a2c234d7ab0e1d7ab to actually see logs from the spring tc test cases, this was done on main already with https://github.com/camunda/zeebe-process-test/commit/6dbc2a30f20dafc783343b71646a5ba921823fff
3. Fixing the spring tc tests by excluding the actuator-autoconfigure dependency via f26795e35385bdce5887a9cddbdaf78d7fe17eee, see this comment on why https://github.com/camunda/zeebe-process-test/pull/1617#issuecomment-2877467613

closes #1389

PR #1669, opened 2025-05-14
ci: use backport-action PAT on git checkout

megglos:
## Description

This is needed so the action can push changes to workflows.

Otherwise backporting workflow changes fail with a permission error like:
```
 ! [remote rejected] backport-1617-to-stable/8.6 -> backport-1617-to-stable/8.6 (refusing to allow a GitHub App to create or update workflow `.github/workflows/build-test.yml` without `workflows` permission)
 ```
see e.g. https://github.com/camunda/zeebe-process-test/actions/runs/15013428320/job/42186200335

Which happened on https://github.com/camunda/zeebe-process-test/pull/1617

Similar change was done on camunda/camunda with https://github.com/camunda/camunda/pull/10785

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -22,6 +22,9 @@ jobs:
       )
     steps:
       - uses: actions/checkout@v4
+        with:
+          # Token for git actions, e.g. git push, needed to be able to e.g. update workflow files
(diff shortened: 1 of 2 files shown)


megglos:
added a commit to fix a recent compile error against camunda/camunda `main`

nicpuppa (approved):

PR #1677, opened 2025-05-20
feat: change slack handle to camunda-ex medic

nicpuppa:
## Description

Changed the slack handle from `zeebe-medic` to `camunda-ex-medic` as ZPT ownership has changed.

Future improvement:

We should maybe change also the slack channel, but for now it's ok


megglos (approved):
Thanks for taking action here!

PR #1686, opened 2025-05-23
ci: new GHA workflow to automate manual release tasks

cmur2:
## Description

This PR introduces a new GHA workflow `create-release.yml` that automatically performs all steps that were previously done manually by the release manager as parts of the user tasks "Upgrade Zeebe version" (in the pom.xml) and "Create release" (on GitHub):

image

image

This change affects 8.3-8.8 since ZPT releases are done identically for all C8 versions.

The goal of this PR is **not** to replace/reengineer the whole `deploy-artifacts` GHA workflow that does the Maven build and upload (too time-consuming compared to value gained) since `deploy-artifacts` works and is also ran `on: push` to `main`.

Instead, this PR only aims to close the gap and automate tasks that are currently manual for smoother release experience (most value). A run of `create-release.yml` will implicitly trigger `deploy-artifacts.yml` at the end.

The new GHA workflow `create-release` can is called from the `zeebe-process-test-release.bpmn` in https://github.com/camunda/zeebe-engineering-processes/pull/557 → removes the need mentioned 2 user tasks.

## Testing

For `create-release` I manually tested by mocking the release branch and vairables to see if the individual steps work and have enough permissions:

* dry-run: https://github.com/camunda/zeebe-process-test/actions/runs/15253189533/job/42894561299
* no dry-run: https://github.com/camunda/zeebe-process-test/actions/runs/15253665687/job/42896122418 triggering https://github.com/camunda/zeebe-process-test/actions/runs/15253679709 (for `deploy-artifacts`)

image

Produced version bump commit: https://github.com/camunda/zeebe-process-test/commit/d5d694e8556dd9dcccbde99363f9ea6098febff9

image

## Related issues

Related https://github.com/camunda/camunda/issues/31509

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -121,7 +121,7 @@ jobs:
         run: |
           shopt -s nocasematch # set matching to case insensitive
           PRE_RELEASE=false
-          if [[ "${RELEASE_VERSION}" =~ ^.*-(alpha|rc|SNAPSHOT)[0-9]+[\.]*[0-9]*$ ]]; then
+          if [[ "${RELEASE_VERSION}" =~ ^.*-(alpha|rc|SNAPSHOT).*$ ]]; then
             PRE_RELEASE=true
           fi
           shopt -u nocasematch # reset it
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,0 +1,188 @@
+---
+name: Create Release
+
+# Used by https://github.com/camunda/zeebe-engineering-processes/blob/main/src/main/resources/release/zeebe-process-test-release.bpmn
+on:
+  workflow_dispatch:
+    inputs:
+      releaseVersion:
+        description: 'releaseVersion: e.g. MAJOR.MINOR.PATCH[-ALPHAn]'
+        type: string
+        required: true
+      minorVersion:
+        description: 'minorVersion: just the MAJOR.MINOR part of the releaseVersion'
+        type: string
+        required: true
+      releaseType:
+        description: 'releaseType: e.g. minor, alpha or patch'
+        type: string
+        required: true
+      nextDevelopmentVersion:
+        description: 'nextDevelopmentVersion: e.g. 8.X.X-SNAPSHOT'
+        type: string
+        required: true
+      isLatest:
+        description: 'isLatest: updates the `latest` docker tag'
+        type: boolean
+        required: false
+        default: false
+      releaseBranch:
+        description: 'releaseBranch: defaults to `release-$releaseVersion` if not set'
+        type: string
+        required: false
+        default: ''
+      dryRun:
+        description: 'dryRun: Whether to perform a dry release where no changes/artifacts are pushed'
+        type: boolean
+        required: true
+        default: false
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    env:
+      RELEASE_BRANCH: ${{ inputs.releaseBranch != '' && inputs.releaseBranch || format('release-{0}', inputs.releaseVersion) }}
+      RELEASE_VERSION: ${{ inputs.releaseVersion }}
+      MINOR_VERSION: ${{ inputs.minorVersion }}
+    steps:
+      - name: Output inputs
+        run: echo "${{ toJSON(github.event.inputs) }}"
+
+      # This step generates a GitHub App token to be used in Git operations as a workaround  for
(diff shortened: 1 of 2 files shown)


megglos (approved):
Nice stuff! 🚀  Looking forward to see it in action on the next releases!

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,188 @@
...
+          git checkout --branch "stable/${MINOR_VERSION}"
+          git push origin "stable/${MINOR_VERSION}"
+
+      # Same logic as in camunda/camunda
+      - name: Determine if pre-release
+        id: pre-release
+        run: |
+          shopt -s nocasematch # set matching to case insensitive
+          PRE_RELEASE=false
+          if [[ "${RELEASE_VERSION}" =~ ^.*-(alpha|rc|SNAPSHOT)[\d]*$ ]]; then
```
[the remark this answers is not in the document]
maxdanilov: or maybe even

```
^.*-(alpha|rc|SNAPSHOT)[0-9]+[\.]*[0-9]*$
```

for cases like 8.8.0-alpha4.1

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,188 @@
...
+          url: ${{ secrets.VAULT_ADDR }}
+          method: approle
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+          exportEnv: false # we rely on step outputs, no need for environment variables
+          secrets: |
+            secret/data/products/camunda/ci/github-actions SLACK_TOPMONOREPORELEASE_WEBHOOK_URL;
+
+      - name: Send failure Slack notification
+        uses: slackapi/slack-github-action@485a9d42d3a73031f12ec201c457e2162c45d02d # v2.0.0
```
maxdanilov, line 165: (suggestion)

Latest version is 2.1.0, we should use it.

maxdanilov (changes requested):
Thank you, leaving 1 blocker and 2 enhancement comments.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,188 @@
...
+# Used by https://github.com/camunda/zeebe-engineering-processes/blob/main/src/main/resources/release/zeebe-process-test-release.bpmn
+on:
+  workflow_dispatch:
+    inputs:
+      releaseVersion:
+        description: 'releaseVersion:'
+        type: string
+        required: true
+      minorVersion:
+        description: 'minorVersion: just the MAJOR.MINOR part of the releaseVersion'
```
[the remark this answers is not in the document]
cmur2: > for reducing human errors, shall we make minorVersion automatically calculated in the workflow from the value of releaseVersion?

It is automatically calculated via a similar expression in the BPMN release process (start event of the zeebe-process-test) so I wouldn't move that calculation around. Humans shouldn't have to interact with it.

image

.github/workflows/create-release.yml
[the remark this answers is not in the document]
cmur2: > for cases like 8.8.0-alpha4.1

🙈 Yea will adjust that!

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,188 @@
...
+          git checkout --branch "stable/${MINOR_VERSION}"
+          git push origin "stable/${MINOR_VERSION}"
+
+      # Same logic as in camunda/camunda
+      - name: Determine if pre-release
+        id: pre-release
+        run: |
+          shopt -s nocasematch # set matching to case insensitive
+          PRE_RELEASE=false
+          if [[ "${RELEASE_VERSION}" =~ ^.*-(alpha|rc|SNAPSHOT)[0-9]+[\.]*[0-9]*$ ]]; then
```
maxdanilov, line 124: for snapshots - the version looks like `8.7.0-SNAPSHOT`, right? then it won't be captured (due to the `[0-9]+` part).

```suggestion
          if [[ "${RELEASE_VERSION}" =~ ^.*-(alpha|rc|SNAPSHOT)[0-9]*[\.]*[0-9]*$ ]]; then
```
or something like it.

maxdanilov (changes requested):

.github/workflows/create-release.yml
maxdanilov: yep, that's fine!

maxdanilov (approved):

.github/workflows/create-release.yml
megglos: actually `.*-SNAPSHOT` would not be a valid release version 🤔 if we do RCs for alphas it would also be rather `*.-alpha2-rc1`. Effectively we could also simply assume any version that has some `-.*` suffix is a pre-release, right?

PR #1691, opened 2025-05-27
ci: add Slack notification when release is done

cmur2:
## Description

`deploy-artifacts` workflow will build and upload Docker and Maven artifacts after being triggered by `create-release` workflow.

The release manager is waiting for the `deploy-artifacts` workflow to finish but there is no visibility in Slack (unlike for the monorepo). This PR adds Slack notifications on the last step both on success and failure so release manager can act.

Follow-up to #1686 

Related to https://github.com/camunda/camunda/issues/31509


megglos (approved):
🚀

cmur2:
@megglos how do I do backports in this repo and which branches do we need to backport to? all from 8.4-8.7?

megglos:
same as on camunda/camunda via the `backport <target-branch>` labels, I went ahead and added all that we still release from + the pending alpha release branch.

PR #1819, opened 2025-07-30
Remove value type loggers test

remcowesterhoud:
## Description

If no value type logger can be found we log an empty string by default. This is good enough as new value types won't be 
supported in ZPT anymore. No need for this test case anymore

## Related issues

closes #

## Definition of Done

Code changes:

Testing:

Documentation:


saig0 (approved):
:+1:

PR #1845, opened 2025-08-26
fix: switch to gRPC endpoints for camunda client

nicpuppa:
## Description

Switch back to gRPC after the default is changed


megglos:
Follow-up on this upstream change https://github.com/camunda/camunda/pull/36992

megglos (approved):

PR #1882, opened 2025-09-18
ci: adjustment for 8.8 minor release

cmur2:
## Description

Since the `stable/8.8` branch already exists, we want to avoid failure in the ZPT release automation when it would try to create it again on the next 8.8 minor release.

## Related issues

Related to https://github.com/camunda/camunda/issues/38074


megglos (approved):
One step ahead!

oleksandr-kriuchenko-lohika (approved):

PR #1941, opened 2025-12-04
test: add simple metrics recorder to test purpose

nicpuppa:
re-add SimpleMetricsRecorder in ZPT

### DIFF
--- a/spring-test/common/src/main/java/io/camunda/zeebe/spring/test/SimpleMetricsRecorder.java
+++ b/spring-test/common/src/main/java/io/camunda/zeebe/spring/test/SimpleMetricsRecorder.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright © 2021 camunda services GmbH (info@camunda.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.camunda.zeebe.spring.test;
+
+import io.camunda.client.metrics.MetricsRecorder;
+import java.util.HashMap;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicLong;
+
+/** Super simple class to record metrics in memory. Typically used for test cases */
+public class SimpleMetricsRecorder implements MetricsRecorder {
+
+  public static final String METRIC_NAME_JOB = "camunda.job.invocations";
+  public static final String ACTION_COMPLETED = "completed";
+  public static final String ACTION_FAILED = "failed";
+  public static final String ACTION_ACTIVATED = "activated";
+  public static final String ACTION_BPMN_ERROR = "bpmn-error";
+
+  public HashMap<String, AtomicLong> counters = new HashMap<>();
+
+  public HashMap<String, Long> timers = new HashMap<>();
+
+  private void increase(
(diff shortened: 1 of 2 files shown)


spring-test/common/src/main/java/io/camunda/zeebe/spring/test/SimpleMetricsRecorder.java
```diff
@@ -0,0 +1,87 @@
...
+  public static final String ACTION_FAILED = "failed";
+  public static final String ACTION_ACTIVATED = "activated";
+  public static final String ACTION_BPMN_ERROR = "bpmn-error";
+
+  public HashMap<String, AtomicLong> counters = new HashMap<>();
+
+  public HashMap<String, Long> timers = new HashMap<>();
+
+  private void increase(
+      final String metricName, final String action, final String type, final int count) {
```
jonathanlukas: There are enums "Name" and "Action" in the Spring Boot starter "JobHandlerMetrics" which can be used instead of a string here.

jonathanlukas (approved):
As it works, I would approve it.

The comments can be discussed as a follow-up.

spring-test/common/src/main/java/io/camunda/zeebe/spring/test/SimpleMetricsRecorder.java
```diff
@@ -0,0 +1,87 @@
...
+
+package io.camunda.zeebe.spring.test;
+
+import io.camunda.client.metrics.MetricsRecorder;
+import java.util.HashMap;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicLong;
+
+/** Super simple class to record metrics in memory. Typically used for test cases */
+public class SimpleMetricsRecorder implements MetricsRecorder {
```
megglos: @jonathanlukas in that regard the upstream change is actually breaking, do you consider it likely any user actually made use of it? 🤔 was the only purpose seemed to have been testing/ZPT or were there other use cases?

megglos (approved):
👍

spring-test/common/src/main/java/io/camunda/zeebe/spring/test/SimpleMetricsRecorder.java
```diff
@@ -0,0 +1,87 @@
...
+      counters.get(key).addAndGet(count);
+    }
+  }
+
+  private String key(final String metricName, final String action, final String type) {
+    final String key = metricName + "#" + action + "#" + type;
+    return key;
+  }
+
+  public long getCount(final String metricName, final String action, final String type) {
```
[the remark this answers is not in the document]
nicpuppa: same as above

spring-test/common/src/main/java/io/camunda/zeebe/spring/test/SimpleMetricsRecorder.java
```diff
@@ -0,0 +1,87 @@
...
+    if (!counters.containsKey(key(metricName, action, type))) {
+      return 0;
+    }
+    return counters.get(key(metricName, action, type)).get();
+  }
+
+  @Override
+  public void increaseActivated(final CounterMetricsContext context) {
+    increase(context.name(), "activated", context.tags().get("type"), context.count());
```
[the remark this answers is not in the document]
nicpuppa: same as above

spring-test/common/src/main/java/io/camunda/zeebe/spring/test/SimpleMetricsRecorder.java
jonathanlukas: Not really. This MetricsRecorder has no use case other than recording metrics and accessing them later in a test. Also, this was never publicly documented.

To be 100% sure, we could add this to CPT as well. @saig0 wdyt?

spring-test/common/src/main/java/io/camunda/zeebe/spring/test/SimpleMetricsRecorder.java
jonathanlukas: lol, on it: https://github.com/camunda/camunda/pull/42073
🤦

PR #1947, opened 2025-12-08
bump testcontainers version

nicpuppa:


megglos (approved):
it can indeed be easy to update to a new major version.

PR #1963, opened 2025-12-17
fix: new rpc methods are not supported by ZPT as the project is deprecated in favour of CPT

nicpuppa:
## Description

New RPC method `evaluateConditional` is not supported in ZPT, as the project is deprecated from 8.8


saig0 (approved):
👍

PR #2307, opened 2026-06-09
fix: actually test (and pass) Spring Boot 4 in CI on stable/8.8

### LINKED ISSUE #2306: Spring Boot 4 CI matrix never tested against Spring Boot 4 (profile "spring-boot-4" not activated)
## Description

The CI "Spring Boot 4" test matrix jobs have **never actually tested against Spring Boot 4**. They silently fall back to the default Spring Boot version (currently `3.5.14`).

During the `8.8.26` release, the deploy run surfaced the warning:

```
[WARNING] The requested profile "spring-boot-4" could not be activated because it does not exist.
```

(seen in [run 27198516460 / job 80303227656](https://github.com/camunda/zeebe-process-test/actions/runs/27198516460/job/80303227656))

### Root cause

The matrix in `.github/workflows/build-test.yml` declares the Spring Boot versions **unquoted**:

```yaml
spring-boot-version: [ 3.5, 4.0]
```

YAML parses the unquoted `4.0` as a **float**, and GitHub Actions renders that float as `4` (the trailing-zero fraction is dropped). The Maven invocation:

```yaml
-P !localBuild,spring-boot-${{ matrix.spring-boot-version }}
```

therefore expands to `-P !localBuild,spring-boot-4`, but the Maven profile is named `spring-boot-4.0` (`spring-test/pom.xml`):

```xml
<profile>
  <id>spring-boot-4.0</id>
  <properties>
    <dependency.spring-boot.version>4.0.1</dependency.spring-boot.version>
  </properties>
</profile>
```

Maven can't find `spring-boot-4`, emits the warning, and the activation is a no-op — so `dependency.spring-boot.version` stays at the default `3.5.14` and the tests run against Spring Boot 3.5, not 4.

Evidence:
- The job is named **"Run tests on ubuntu-latest and Spring-Boot 4"** (note `4`, not `4.0`) — confirming the matrix value rendered as `4`.
- The logged command shows `-P !localBuild,spring-boot-4`.
- Two `could not be activated` warnings appear in the job log.

The `3.5` entry is unaffected: it is an intentional placeholder for the default version (there is no `spring-boot-3.5` profile by design), and a float with a non-zero fraction renders unchanged.

### Expected behaviour

The "Spring Boot 4" matrix jobs should activate the `spring-boot-4.0` profile and genuinely resolve/test against Spring Boot `4.0.1`, with no "could not be activated" warning.

### Affected

This was introduced in #2007 (profile + workflow added together). The defect is present on at least:
- `main`
- `stable/8.8`

Both define the `spring-boot-4.0` profile and use the unquoted `4.0` matrix value, so both are affected. Other active stable branches with the same setup should be checked too.

## Proposed fix

Quote the matrix values so YAML keeps them as strings:

```yaml
spring-boot-version: [ '3.5', '4.0' ]
```

(both occurrences — the unit-test matrix and the testcontainers matrix). This makes `${{ matrix.spring-boot-version }}` expand to `4.0`, matching the `spring-boot-4.0` profile id.

Alternatively, rename the profile to `spring-boot-4` to match the coerced value — but quoting the matrix is the more robust fix and keeps the profile id semantically versioned.

## Verification

After the fix, the "Spring Boot 4" jobs should:
- run `-P !localBuild,spring-boot-4.0`,
- show **no** "could not be activated" warning,
- and resolve `spring-boot-dependencies:4.0.1` (verifiable via `mvn dependency:tree` / the effective Spring Boot version in the build).

## Environment

- Component: `zeebe-process-test` CI (`.github/workflows/build-test.yml`) + `spring-test/pom.xml`
- Affected versions: `main` (`8.8.27-SNAPSHOT`), `stable/8.8`

megglos:
## Description

Fixes #2306 and makes the Spring Boot 4 CI matrix on `stable/8.8` **actually test against — and pass on — Spring Boot 4**.

### 1. The original bug (silent no-op)
The matrix value `4.0` was unquoted, so YAML parsed it as a float rendered as `4`. The Maven invocation expanded to `-P !localBuild,spring-boot-4`, but the profile is `spring-boot-4.0`, so it never activated (`[WARNING] ... could not be activated`) and the build silently fell back to the default Spring Boot 3.5.14. **Spring Boot 4 was never tested.** Fixed by quoting the matrix values (both matrices).

### 2. What quoting exposed — two real SB4 incompatibilities
Once the profile actually activated, the SB4 jobs failed, revealing genuine (previously-masked) issues:

- **JUnit version.** The root pom imports `junit-bom` (5.13.4) *before* `spring-boot-dependencies`, so its version wins. Spring Boot 4 / Spring Framework 7 require **JUnit 6** — `SpringExtension` calls `ExtensionContext.Store.computeIfAbsent(...)`, absent in JUnit 5 → `NoSuchMethodError`. Fixed by aligning `dependency.junit.version` with the version `spring-boot-dependencies:4.0.1` manages (**6.0.1**), scoped to the profile.

- **Camunda Spring SDK starter.** `camunda-spring-boot-starter` targets Spring Boot 3; under SB4 its `CamundaActuatorConfiguration.micrometerMetricsRecorder` fails with `@ConditionalOnMissingBean ... BeanTypeDeductionException`. The SDK ships a dedicated **`camunda-spring-boot-4-starter`** (published for 8.8.x). Selected via a `camunda.spring-boot.starter.artifactId` property that the `spring-boot-4.0` profile overrides; the default build is unchanged.

## Verification (local, `-P spring-boot-4.0`)
- Embedded module test suite **passes**, including the previously-failing `JobWorkerAnnotationOnTestClassTest` and `ZeebeClientDisabledInCamundaSpringTest`.
- Dependency resolution confirmed: SB4 profile → `camunda-spring-boot-4-starter` + JUnit 6.0.1 (no duplicate starters); default build → `camunda-spring-boot-starter` + JUnit 5.13.4.
- The testcontainers SB4 job needs Docker and is verified by CI here.

## Companion docs PR
User-facing guidance for consuming `spring-boot-starter-camunda-test` under Spring Boot 4 (exclude the SB3 starter, add `camunda-spring-boot-4-starter`): **camunda/camunda-docs#9040** (covers 8.8 and 8.7).

## Scope
Only `stable/8.8` was affected by the original bug. `main`/`stable/8.9` already default to Spring Boot 4.0.6 (their 8.9.x SDK is SB4-native, so they use the regular starter); `stable/8.7`/`stable/8.6` use `3.5` which YAML preserves correctly. See the [branch assessment on #2306](https://github.com/camunda/zeebe-process-test/issues/2306#issuecomment-4660121170).

Closes #2306

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -71,7 +71,10 @@ jobs:
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         # Note: `3.5` is an invalid profile just serving as a placeholder for the default version.
-        spring-boot-version: [ 3.5, 4.0]
+        # Values MUST be quoted: unquoted `4.0` is parsed as a YAML float and rendered as `4`,
+        # so `spring-boot-${{ matrix.spring-boot-version }}` would resolve to the non-existent
+        # `spring-boot-4` profile instead of `spring-boot-4.0` (silently testing the default version).
+        spring-boot-version: [ '3.5', '4.0' ]
 
     steps:
       - name: Checkout
@@ -129,7 +132,9 @@ jobs:
       max-parallel: 3
       matrix:
         # Note: `default` is an invalid profile just serving as a placeholder with no effect.
-        spring-boot-version: [ default, 4.0 ]
+        # Quote version values: unquoted `4.0` is parsed as a YAML float and rendered as `4`,
+        # which would resolve to the non-existent `spring-boot-4` profile instead of `spring-boot-4.0`.
+        spring-boot-version: [ default, '4.0' ]
     env:
       IMAGE_NAME: "qa-tests"
 
--- a/pom.xml
+++ b/pom.xml
@@ -214,6 +214,14 @@
         <version>${dependency.zeebe.version}</version>
       </dependency>
 
+      <!-- Spring Boot 4 variant of the Camunda Spring SDK starter, selected via the
+           spring-boot-4.0 profile (see spring-test modules). -->
+      <dependency>
+        <groupId>io.camunda</groupId>
+        <artifactId>camunda-spring-boot-4-starter</artifactId>
+        <version>${dependency.zeebe.version}</version>
+      </dependency>
+
       <dependency>
         <groupId>io.camunda</groupId>
         <artifactId>camunda-client-java</artifactId>
--- a/spring-test/common/pom.xml
+++ b/spring-test/common/pom.xml
@@ -41,7 +41,7 @@
     </dependency>
     <dependency>
       <groupId>io.camunda</groupId>
-      <artifactId>camunda-spring-boot-starter</artifactId>
+      <artifactId>${camunda.spring-boot.starter.artifactId}</artifactId>
     </dependency>
     <dependency>
       <groupId>io.micrometer</groupId>
--- a/spring-test/embedded/pom.xml
+++ b/spring-test/embedded/pom.xml
@@ -46,7 +46,7 @@
 
     <dependency>
       <groupId>io.camunda</groupId>
-      <artifactId>camunda-spring-boot-starter</artifactId>
+      <artifactId>${camunda.spring-boot.starter.artifactId}</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/spring-test/pom.xml
+++ b/spring-test/pom.xml
@@ -22,6 +22,9 @@
   </modules>
 
   <properties>
+    <!-- Camunda Spring SDK starter artifact; swapped to the Spring Boot 4 variant
+         by the spring-boot-4.0 profile below. -->
+    <camunda.spring-boot.starter.artifactId>camunda-spring-boot-starter</camunda.spring-boot.starter.artifactId>
     <plugin.version.license>5.0.0</plugin.version.license>
     <version.java>17</version.java>
   </properties>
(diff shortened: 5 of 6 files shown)


saig0:
The CI jobs fail with the error message:

```
Error:  Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.001 s <<< FAILURE! -- in io.camunda.zeebe.spring.client.jobhandling.SmokeTest
Error:  io.camunda.zeebe.spring.client.jobhandling.SmokeTest -- Time elapsed: 0.001 s <<< ERROR!
java.lang.NoSuchMethodError: 'java.lang.Object org.junit.jupiter.api.extension.ExtensionContext$Store.computeIfAbsent(java.lang.Object, java.util.function.Function, java.lang.Class)'
	at org.springframework.test.context.junit.jupiter.SpringExtension.getTestContextManager(SpringExtension.java:429)
	at org.springframework.test.context.junit.jupiter.SpringExtension.beforeAll(SpringExtension.java:172)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	Suppressed: java.lang.NoSuchMethodError: 'java.lang.Object org.junit.jupiter.api.extension.ExtensionContext$Store.computeIfAbsent(java.lang.Object, java.util.function.Function, java.lang.Class)'
		at org.springframework.test.context.junit.jupiter.SpringExtension.getTestContextManager(SpringExtension.java:429)
		at org.springframework.test.context.junit.jupiter.SpringExtension.afterAll(SpringExtension.java:183)
		... 1 more
```

megglos:
@saig0 apologies for the noise, this should have been a draft until ready

saig0 (approved):
👍

PR #2326, opened 2026-06-17
fix: move deployment ID extraction to notify-release job

szpraat:
## Problem

The "Extract Maven Central deployment ID" step was inside the `deploy` job and called `GET /actions/jobs/{job_id}/logs` on **itself while still running**. GitHub's API returns an incomplete log archive for in-progress jobs — the `deploymentId:` line written by the Maven publish step was never captured, so the step always emitted `not-found`.

Evidence from the 8.9.9 release run ([job 81707178768](https://github.com/camunda/zeebe-process-test/actions/runs/27630933385/job/81707178768)):
- `16:20:03` — Maven logged `deploymentId: 45647a3c-fd36-4719-a226-6ce5d5d55cd4` ✅
- `16:21:58` — extraction step fetched own job logs via API
- `16:22:00` — `"zpt_maven_central_deployment_id": "not-found"` ❌

The same pattern works in `camunda/camunda` because there the extraction runs in a **downstream notification job** that `needs` the Maven Release job — the target job is fully completed before the API is queried.

## Fix

Move the step to `notify-release`, which already `needs: [deploy]`. By the time `notify-release` runs, `deploy` is done and the GitHub API returns its finalized log archive.

Changes:
- Remove extraction step from `deploy` job
- Remove `outputs.deploymentId` from `deploy` job (no longer needed)
- Remove `actions: read` from `deploy` permissions (no longer needed)
- Add extraction step as first step in `notify-release`
- Add `actions: read` to `notify-release` permissions
- Read deployment ID from `steps.extract-deployment-id.outputs.deploymentId` (local step) instead of `needs.deploy.outputs.deploymentId`

### DIFF
--- a/.github/workflows/deploy-artifact.yml
+++ b/.github/workflows/deploy-artifact.yml
@@ -31,10 +31,6 @@ jobs:
     if: ${{ always() && (needs.test.result == 'success' || needs.test.result == 'skipped') }}
     permissions:
       contents: write
-      actions: read
-
-    outputs:
-      deploymentId: ${{ steps.extract-deployment-id.outputs.deploymentId }}
 
     env:
       DRY_RUN: ${{ github.event_name == 'workflow_dispatch' }}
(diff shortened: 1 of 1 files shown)


saig0 (approved):
👍

